PR #4, opened 2021-08-04
drafted changes to sombrero wrapper

mmesiti:
Some things to discuss:
1. what needs to be inside `SombreroBenchmark.__init__` and what does not?
2. sombrero is not able to make use of multithreading at the moment
   (it is not even compiled with `-fopenmp`), so `OMP_NUM_THREADS` has no effect
   (last time I tried activating openMP it made it slower than the MPI-only option)
3. I set the reference flops to 0 in the base class,
   mimicking what was done for the castep benchmark.
4. I added benchmarks for the strong scaling on the medium lattice 
   and the weak scaling on the medium lattice - this choice was pretty arbitrary.
   We need to discuss how many cases we want to run.
   Also, we use
   ```
   @rfm.parameterized_test(*scaling_config())
   ```
   but `scaling_config()` is shared between all the benchmarks.
   In principle there can be some cases we don't want to run
   with Sombrero in all the sizes (small, medium, large and very large)
   compared to other benchmarks - for that we might create 
   a `sombrero_filter_scaling_config(problem_size)` function 
   that filters out the unrealistic cases?
   Poking @edbennett for an educated opinion.

### DIFF
--- a/apps/sombrero/reframe_sombrero.py
+++ b/apps/sombrero/reframe_sombrero.py
@@ -5,38 +5,67 @@
 
 import reframe as rfm
 import reframe.utility.sanity as sn
+from modules.reframe_extras import scaling_config
 
 
 @rfm.simple_test
 class SombreroBenchmark(rfm.RegressionTest):
-    valid_systems = ['*']
-    valid_prog_environs = ['*']
-    build_system = 'Spack'
-    executable = 'sombrero.sh'
-    executable_opts = ['-w', '-s', 'medium']
-    num_tasks = 4
-    time_limit = '10m'
-    variables = {
-        'OMP_NUM_THREADS': '4',
-        'OMP_PLACES': 'cores'
-    }
-    reference = {
-        '*': {
-            'flops': (30, -0.5, 0.5, 'Gflops/seconds'),
+    def __init__(self):
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+        self.build_system = 'Spack'
+        self.executable = 'sombrero.sh'
+        self.num_tasks = 2
+        self.time_limit = '10m'
+        self.reference = {
+            '*': {
+                'flops_average': (0, None, None, 'Gflops/seconds'),
+                'flops_1': (0, None, None, 'Gflops/seconds'),
+                'flops_2': (0, None, None, 'Gflops/seconds'),
+                'flops_3': (0, None, None, 'Gflops/seconds'),
+                'flops_4': (0, None, None, 'Gflops/seconds'),
+                'flops_5': (0, None, None, 'Gflops/seconds'),
+                'flops_6': (0, None, None, 'Gflops/seconds'),
+            }
         }
-    }
 
     @run_before('compile')
     def setup_build_system(self):
-        self.build_system.specs = ['sombrero@1.0']
+        self.build_system.environment = '.'
+        self.build_system.specs = ['sombrero@2021-07-31']
 
     @run_before('sanity')
     def set_sanity_patterns(self):
-        self.sanity_patterns = sn.assert_found(r'\[RESULT\] SUM', self.stdout)
+        self.sanity_patterns = sn.all(
+            [sn.assert_found(r'\[RESULT\] SUM', self.stdout)] + [
+                sn.assert_found(r'\[RESULT\]\[0]\ Case '
+                                f'{i}', self.stdout) for i in range(1, 7)
+            ])
 
     @run_before('performance')
     def set_perf_patterns(self):
-        self.perf_patterns = {
-            'flops': sn.extractsingle(r'\[RESULT\] SUM (\S+) Gflops/seconds',
-                                      self.stdout, 1, float),
-        }
+
+        pattern_template = r'\[RESULT\]\[0\] Case {i} (\S+) Gflops/seconds'
+
+        pattern_dict = [(f"flops_{i}", pattern_template.format(i=i))
+                        for i in range(1, 7)]
+        pattern_dict.append(
+            ('flops_average', r'\[RESULT\] SUM (\S+) Gflops/seconds'))
+
+        self.perf_patterns = dict(
+            (pattern[0], sn.extractsingle(pattern[1], self.stdout, 1, float))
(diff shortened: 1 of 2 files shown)


edbennett:
Yes, the strong scaling cases each have a maximum number of ranks that can be used. I'm not sure how useful `max_node_factor` and `core_factor` are here since the optimal number of cores per node is not a fixed fraction of the node (e.g. a 32-core node, 32 is usually the ideal number; on a 40-core node, 36 may be). Probably having an application-specific filter is the way to go, rather than overriding `scaling_config` altogether.

apps/sombrero/reframe_sombrero.py
```diff
@@ -5,38 +5,67 @@
...
-    }
-    reference = {
-        '*': {
-            'flops': (30, -0.5, 0.5, 'Gflops/seconds'),
+    def __init__(self):
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+        self.build_system = 'Spack'
+        self.executable = 'sombrero.sh'
+        self.num_tasks = 2
```
giordano, line 18: `num_tasks` is a bit tricky.  ReFrame uses this parameter to set up the scheduler/launcher with the appropriate number of MPI tasks, so that the job script would have something like
```sh
mpirun -n $num_tasks sombrero.sh ...
```
But the `sombrero.sh` shell driver internally runs the other `sombrero` applications binaries with `mpirun -n ...`, so that there would be two nested `mpirun` calls, which OpenMPI didn't like at all.  That's why in https://github.com/ukri-excalibur/excalibur-tests/commit/a2981df0c6d7325a3ef4138139c8bfd39fd1e9f0 I didn't set `num_tasks` (so it defaults to 1 and the `mpirun` launcher isn't used) and only pass the `-n` argument to `sombrero.sh`

edbennett: In principle `sombrero.sh` is just a convenience wrapper for the actual tests `sombrero1` through `sombrero6`; there's nothing stopping us directly running the individual tests, which would presumably avoid the double `mpirun` call. (The `SUM` calculation would need to be re-done explicitly, but that doesn't seem particularly onerous.)

mmesiti: We can make a quick change to the install step in the spack package to make  `sombrero{1..6}` available in the `$PATH`

giordano: Yes, that'd make it easier to call those commands directly :slightly_smiling_face:

giordano:
Ok, I've dome some changes in the `main` branch.  In particular, I've added to the Sombrero benchmark what I think is a good solution for the Spack environment (I discussed this with Michele last week):
https://github.com/ukri-excalibur/excalibur-tests/blob/1a1c6864263efc424eac7252b0c85547092f5af4/apps/sombrero/reframe_sombrero.py#L29-L46
This is now also explained in the [README](https://github.com/ukri-excalibur/excalibur-tests/blob/1a1c6864263efc424eac7252b0c85547092f5af4/README.md#spack).  This means this pull request doesn't need to provide a Spack environment, it'd be the same as the one automatically created by ReFrame as last fallback option.  However this requires ReFrame 3.7.3 (which is not released yet, so at the moment you'd need to check out latest `master` of the Git repository).

Regarding MPI, I believe the best solutions are

* either not hardcode `mpirun -n ...` inside `sombrero.sh`,
* or don't use `sombrero.sh` at all, but this is tricky because as far as I understand each benchmark takes has single exacutable (so we'd need 6 different benchmark files for Sombrero?).  But I'm not 100% sure about this, I might be missing something in ReFrame

As it is now, the problem is that if we don't set `num_tasks` at all, ReFrame will not set up the job to require multiple MPI slots, and MPI won't be happy if you then try to run `mpirun -n 4`.  However, if we do set `num_tasks`, ReFrame will also prepend `mpirun -n ...` to `sombrero.sh`, so that there will be two nested `mpirun` calls.

It'd be great if you could rebase this pull request on latest `main` to adapt to the changes I made to `reframe_sombrero.py` :slightly_smiling_face:

mmesiti:
I have merged your changes, and made the tests parametrized using `parameter` 
instead of the `reframe.parametrized_test()` decorator, as recommended.

A modification to make SOMBRERO's 6 tests available in the $PATH 
is now in a pull request: https://github.com/spack/spack/pull/25421
I have not managed yet to run test my changes end-to-end
(it currently fails with an internal error by reframe on my machine). 
Investigating also on other machines.

apps/sombrero/src/spack.yaml
```diff
@@ -0,0 +1,4 @@
+spack:
```
giordano, line 1: This file is no more needed and can be removed

apps/sombrero/reframe_sombrero.py
```diff
@@ -7,25 +7,102 @@
...
-    valid_systems = ['*']
-    valid_prog_environs = ['*']
-    build_system = 'Spack'
-    executable = 'sombrero.sh'
-    executable_opts = ['-w', '-s', 'medium', '-n', '4']
-    time_limit = '10m'
-    reference = {
-        '*': {
-            'flops': (30, -0.5, 0.5, 'Gflops/seconds'),
+    params = parameter(SombreroCase.generate(scaling_config))
```
giordano, line 80: I'm a bit lost: where does `parameter` come from?

giordano (commented):
It looks overall good.  It'd be great to have a README file (see for example [castep](https://github.com/ukri-excalibur/excalibur-tests/blob/c7f5079a561dbbc68d86857aa5f62dc6dc882804/apps/castep/README.md)) to explain what the tests do and how to run them, since I believe one needs to pass extra flags to select the specific case?

apps/sombrero/reframe_sombrero.py
mmesiti: `parameter` is a built in type of reframe:
https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html?highlight=parameter#built-in-types

giordano (changes requested):
Can you please rebase on latest `main`?  All changes for the cosma environment aren't needed here.

PR #7, opened 2021-08-25
Setup for cosma8

mmesiti:
This change works on the `sombrero-draft` branch.
(submits to compute nodes using account tc004)
Reapplied it to `main` as it should be orthogonal,
but needs verification.


giordano:
Thanks!

PR #14, opened 2021-11-09
Update Intel MPI Benchmarks

tkoskela:
Adds spack build for `apps/imb` and replaces outdated code in parameterized tests. This test runs from `IMB-MPI1`
- `PingPong` on 2 nodes with 1 rank/node, 
- `Uniband` on 1,2,4,...,256 ranks with default pinning
- `Biband` on 1,2,4,...,256 ranks with default pinning


spack-environments/csd3/spack.yaml
```diff
@@ -4,7 +4,7 @@
 # configuration settings.
 spack:
   # add package specs to the `specs` list
-  specs: []
+  specs: [sombrero@2021-08-16, intel-mpi-benchmarks@2019.6]
```
tkoskela, line 7: These must have been updated automatically by spack. Should I remove them?

reframe_config.py
```diff
@@ -13,16 +13,20 @@
                     'descr': 'CSD3 Login nodes',
                     'scheduler': 'local',
                     'launcher': 'local',
-                    'environs': ['gnu'],
+                    'environs': ['gnu','intel'],
```
giordano, line 16: I don't think we need to add ReFrame programming environments: we aren't planning to use ReFrame's own build system at any point, Spack will deal with all of that, and that's independent from these programming environments.

spack-environments/csd3/spack.yaml
giordano: Yeah, I think that should be better.  I'm still not 100% sure about what this though: I'd like to keep it empty by default in case we end up having incompatible packages from different benchmarks, but hopefully that won't happen.

apps/imb/reframe_imb.py
```diff
@@ -0,0 +1,119 @@
...
+from modules.reframe_extras import ScalingTest
+from modules.utils import identify_build_environment
+
+Metric = namedtuple('Metric', ['column', 'function', 'unit', 'label'])
+
+class IMB_base(rfm.RegressionTest):
+    METRICS = []
+
+    def __init__(self):
+        self.valid_systems = ['*:compute-node']
```
giordano, line 28: Side note, I'm thinking that we may want to remove the login node partitions from the systems.  They're useful for _quick_ tests, but not much more than that.

apps/imb/reframe_imb.py
```diff
@@ -0,0 +1,119 @@
...
+        self.valid_prog_environs = ['intel']
```
giordano, line 29: Related to the comment in `reframe_config.py`: I think this doesn't have any effect when the Spack build system is used, does it?  If you want to force Intel compilers you should use below a Spack spec like `intel-mpi-benchmarks@2019.6%intel`, and `intel-mpi-benchmarks@2019.6%intel^intel-mpi` if you also want to use Intel MPI.

apps/imb/reframe_imb.py
tkoskela: I'm inclined to agree, we shouldn't really be setting up things to run on login nodes

apps/imb/reframe_imb.py
tkoskela: Ok, I thought this programming environment would get somehow passed down to spack, but it seems to run happily in the `gnu` environment as well.

reframe_config.py
```diff
@@ -107,6 +111,12 @@
             'cxx': 'g++',
             'ftn': 'gfortran'
         },
+        {
+            'name': 'intel',
+            'cc': 'icc',
+            'cxx': 'icpc',
+            'ftn': 'ifort'
+        },
```
giordano, line 119: I think we can also remove the programming environment altogether.  Spack is not going to consider it anyways.

spack-environments/csd3/spack.yaml
```diff
@@ -101,9 +101,51 @@ spack:
...
+      flags: {}
+      operating_system: scientific7
+      target: x86_64
+      modules: []
+      environment: {}
+      extra_rpaths: []
   packages:
     openmpi:
       externals:
+      - spec: openmpi@4.0.0~cuda+cxx~cxx_exceptions~java~memchecker+pmi+pmix~sqlite3+static~thread_multiple~wrapper-rpath
```
giordano, line 146: If you rebase on `main` you'll get also `intel-mpi` from the system, even though that's probably not going to be used anyways unless you force this MPI provider in the Spack spec with `^intel-mpi`

giordano:
Also, the REAMDE.md of this directory should probably be updated

reframe_config.py
tkoskela: Why don't we just use the`builtin` environment if we never use any compilers?

giordano: I seem to remember trying to remove all environments from here and getting an error because at least one programming environment must be specified, and that's why I had left the "gnu" one above, not because it was needed for something specific.  Is there a way to specify a builtin environment?  I hadn't explored that

spack-environments/csd3/spack.yaml
tkoskela: I don't have strong feelings over whether I should request `intel-mpi` or not. I guess ideally that should be a `parameter()`

tkoskela: It is now a parameter.

tkoskela:
There were some. I may have added one or two. The Intel MPI benchmarks are huge, this only calls a couple of modes.

PR #19, opened 2022-03-21
Add a base test class and new logging format

t-young31:
This PR includes a suggestion for a logging format along with a base test class, which all the ReFrame benchmarks could/should inherit from.

***
**Notes**

- The use of logged reframe variables (e.g. `x = variable(int, loggable=True)`) currently requires ReFrame [v3.11.0rc1](https://github.com/eth-cscs/reframe/releases/tag/v3.11.0-rc1), which is not pip-installable. Hopefully this feature will make its way into a properly released version soon. 
- Currently no benchmarks inherit from the base class (see https://github.com/t-young31/excalibur-tests/blob/main/apps/gromacs/benchmark.py for a full example)
- Scaling tests/benchmarks would require a something like [this](https://github.com/t-young31/excalibur-tests/blob/4f237b398772c98f3076698b067173bf9a16b510/modules/base_test.py#L65), as a _parameter_ cannot be a pre-defined _variable_.

Please feel free to add comments/suggestions or push to this branch 😄

### DIFF
--- a/modules/base_test.py
+++ b/modules/base_test.py
@@ -1,0 +1,69 @@
+"""
+Base class definitions to enable consistent logging. Requires ReFrame v. 3.11.x
+"""
+import math
+import reframe as rfm
+from reframe.core.decorators import run_after
+
+
+class DiRACTest(rfm.RegressionTest):
+
+    num_total_cores = variable(int, loggable=True)
+    num_omp_threads = variable(int, loggable=True)
+    num_mpi_tasks = variable(int, loggable=True)
+    num_mpi_tasks_per_node = variable(int, loggable=True)
+    num_nodes = variable(int, loggable=True)
+
+    @run_after('setup')
+    def set_attributes_after_setup(self):
+        """Set the required MPI and OMP ranks/tasks/threads"""
+
+        self.num_mpi_tasks = self.num_tasks = max(self.num_total_cores//self.num_omp_threads, 1)
+
+        try:
+            cpus_per_node = self._current_partition.processor.num_cpus
+            if cpus_per_node is None:
+                raise AttributeError('Cannot determine the number of cores PP')
+
+            self.num_nodes = math.ceil(self.num_mpi_tasks / cpus_per_node)
+
+        except AttributeError:
+            print('WARNING: Failed to determine the number of nodes required '
+                  'defaulting to 1')
+            self.num_nodes = 1
+
+        self.num_mpi_tasks_per_node = math.ceil(self.num_mpi_tasks / self.num_nodes)
+        self.num_tasks_per_node = self.num_mpi_tasks_per_node
+
+        if self.num_total_cores // self.num_omp_threads == 0:
+            print('WARNING: Had fewer total number of cores than the default '
+                  f'number of OMP threads, using {self.num_total_cores} OMP '
+                  f'threads')
+            self.num_omp_threads = self.num_total_cores
+
+        self.num_cpus_per_task = self.num_omp_threads
+        self.variables = {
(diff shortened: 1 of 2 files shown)


modules/base_test.py
```diff
@@ -0,0 +1,69 @@
...
+from reframe.core.decorators import run_after
+
+
+class DiRACTest(rfm.RegressionTest):
+
+    num_total_cores = variable(int, loggable=True)
+    num_omp_threads = variable(int, loggable=True)
+    num_mpi_tasks = variable(int, loggable=True)
+    num_mpi_tasks_per_node = variable(int, loggable=True)
+    num_nodes = variable(int, loggable=True)
```
giordano, line 15: I think most if not all of these variables are standard ReFrame variables, and it'd be good to avoid duplication:

* `num_omp_threads` is [`num_cpus_per_task`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.num_cpus_per_task)
* `num_mpi_tasks` is [`num_tasks`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.num_tasks) (this is automatically used as argument to `mpiexec` to specify the number of MPI processes)
* `num_mpi_tasks_per_node` is [`num_tasks_per_node`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.num_tasks_per_node)

`num_total_cores` and `num_nodes` are useful, but I think those can be inferred from other similar existing variables (which may still need to be added to the log)

modules/base_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @run_after('setup')
+    def set_attributes_after_setup(self):
+        """Set the required MPI and OMP ranks/tasks/threads"""
+
+        self.num_mpi_tasks = self.num_tasks = max(self.num_total_cores//self.num_omp_threads, 1)
+
+        try:
+            num_cores_per_node = self._current_partition.processor.num_cpus
```
giordano, line 24: `self._current_partition` looks an internal attribute?  The user-facing one is [`self.current_partition`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html?#reframe.core.pipeline.RegressionTest.current_partition).  I don't think accessing this should ever result in a `AttributeError`.

t-young31: This is an excellent point – I'm sure there was a reason for accessing the private attr but it escapes me now. If it works with the public one then feel free to change

Something in that try definitely did, but this was long ago and probably with a different ReFrame version so 🤷

reframe_config.py
```diff
@@ -185,13 +185,20 @@
...
                     'format': (
-                        '%(check_job_completion_time)s|reframe %(version)s|'
-                        '%(check_info)s|jobid=%(check_jobid)s|'
+                        '%(check_job_completion_time)s|'
+                        'reframe %(version)s|'
+                        '%(check_info)s|'
+                        'jobid=%(check_jobid)s|'
                         '%(check_perf_var)s=%(check_perf_value)s|'
-                        'ref=%(check_perf_ref)s '
+                        'units=%(check_perf_unit)s|'
```
giordano, line 193: What's the rationale for replacing `%(check_perf_unit)s` with `units=%(check_perf_unit)s`?  Is it for making it easier to identify the field?  I think that's ok, but asking to understand the reason of the change.  Although, I'd keep it closer to the rest of the reference values.

t-young31: > Is it for making it easier to identify the field? 

Yep, exactly that. Happy for it to go wherever!

tkoskela:
Closing as obsolete

PR #24, opened 2022-06-13
Add more fields to the perflog format

giordano:
Taking the format from #19, without the base class (for the time being).  I only removed/renamed variables as per [this comment](https://github.com/ukri-excalibur/excalibur-tests/pull/19/commits/4f237b398772c98f3076698b067173bf9a16b510#r874654438) to stick to standard variables.  CC: @t-young31


t-young31 (approved):
Looks great 👍🏼

PR #25, opened 2022-06-22
Add data-intensive benchmarks for DiRac

giordano:
CC: @lkr-ind @t-young31


apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The repository contains the following directories and files:-
+
+1.  **Additional_Input_Data:-** This folder contains instructions to get the input data required for Ramses. We do not need any other input data for other benchmarks. We cannot include them in the current repository as they are big in size.
+2. **Executables:-** This directory contains the binary files or the executable based on the system name. 
+3.  **graphing:-** This directory contains the code for post-processing or collection of results from the perflogs directory (contains the performance log and the output we want from the benchmarks).
+4. **Readme.md:-** File explaining the directory structure and instructions to run the code.
+5. **Settings.py:-** This file contains various environments and module settings which are used by Reframe to generate the submission scripts.
+6. **Others:-** All other directories are the apps/benchmarks that we would like to run. Each of the app contains the following files:-
+   1. **perflogs:-** Also known as performance logs. This directory contains that output that we get by running the benchmark on a particular system.
```
t-young31, line 27: having the perflogs scattered is perhaps not ideal? could all be in /perflogs

apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+   2. **Pyhton code:-** There is one file with extension `.py` which is the reframe code to generate and run the benchmark.
+   3. **SRC:-** This is the directory which contains the various inputs for the test cases under consideration. It is this directory where you would need to copy the executable from the `Executable` directory to run the benchmark. Please read the instructions below to know what needs to be copied and where.
+
+### Steps to run a benchmark
+
+1. You would first need to use the settings file that is included in the current repository. This file has been specifically written for Dial3 cluster while the logging format at the end is generic and can be adapted to any system.
```
t-young31, line 33: needs revising to use /reframe_config.py

apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+   cp Executables/Dial3/Ramses/ramses3d Ramses/src/
+   ```
+   
+   In each of the app that you would like to run, you will see a python code (Based on Reframe) in which we defined the executable name. For example, in case on Ramses, you will notice a file `ramses.py` in `Bmark_reframe/Ramses` directory. If you open this file in your favourite editor, you will find the following instruction in the first few lines of the code,
+   
+   ```python
+   self.executable = './ramses3d'
+   ```
+
+   Make sure that the executable you copied and the executable name in Python code are same. At present, the files in the `Executable` directory may have different names. This is for our reference as it helps us to know which compiler was used to generate the binary file and which environment we need to choose.
```
t-young31, line 52: can be removed in the absence of an Executable/ directory

apps/dirac/Ramses/ramses.py
```diff
@@ -0,0 +1,94 @@
...
+#------------------------------------------------------------------------------------------------------------------------------------
+class RamsesMPI(rfm.RunOnlyRegressionTest):
+    
+    def __init__(self):
+
+        self.time_limit = '0d0h10m0s'
+        self.exclusive_access=True
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+        self.executable = './ramses3d'
```
t-young31, line 16: Not for this PR, but in the base class it might be nice to have an assert that this exists before running a benchmark (unless of course ReFrame does this already)

apps/dirac/Additional_Input_Data/readme.txt
```diff
@@ -0,0 +1,14 @@
+The input files intended for this folder are too large for a github repository.
+These files can however be downloaded from the University of Edinburgh's DataSync service.
+
+Login to a local machine that has an internet browser.
+Browse to "https://datasync.ed.ac.uk/index.php/s/ju6knXo5TVchspd", the password is "dirac3bm".
```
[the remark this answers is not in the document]
lkr-ind: Yes, this is meant to be Public.

apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+At present, the build step is separate and is not integrated with the Reframe. For building codes, we can use either Spack and its associated settings file or the normal build process using make. Moreover, the source codes are not public and hence they have not been include in this repository.
+
+However, for 3 DiRAC systems namely: 
+
+1. DIaL3 (University of Leicester), 
+2. Cosma8 (University of Durham) and,
+3. CSD3 Icelake nodes (University of Cambridge),
+
+we have included the executable in the directory `Bmark_reframe/Executables/System_name` where `System_name` can be any of the 3 system names mentioned above.
```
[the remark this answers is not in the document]
lkr-ind: This has been addressed in PR#26.

apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+1. DIaL3 (University of Leicester), 
+2. Cosma8 (University of Durham) and,
+3. CSD3 Icelake nodes (University of Cambridge),
+
+we have included the executable in the directory `Bmark_reframe/Executables/System_name` where `System_name` can be any of the 3 system names mentioned above.     
+
+We plan to add the instructions to get the SRC code in near future.
+
+### Organisation of the Repository
```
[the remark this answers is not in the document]
lkr-ind: The suggestion has been incorporated in #26.

apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The repository contains the following directories and files:-
+
+1.  **Additional_Input_Data:-** This folder contains instructions to get the input data required for Ramses. We do not need any other input data for other benchmarks. We cannot include them in the current repository as they are big in size.
+2. **Executables:-** This directory contains the binary files or the executable based on the system name.
```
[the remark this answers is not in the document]
lkr-ind: Yes, you are right. Now changed the text in #26

apps/dirac/README.md
```diff
@@ -0,0 +1,117 @@
...
+We plan to add the instructions to get the SRC code in near future.
+
+### Organisation of the Repository
+
+The repository contains the following directories and files:-
+
+1.  **Additional_Input_Data:-** This folder contains instructions to get the input data required for Ramses. We do not need any other input data for other benchmarks. We cannot include them in the current repository as they are big in size.
+2. **Executables:-** This directory contains the binary files or the executable based on the system name. 
+3.  **graphing:-** This directory contains the code for post-processing or collection of results from the perflogs directory (contains the performance log and the output we want from the benchmarks).
+4. **Readme.md:-** File explaining the directory structure and instructions to run the code.
```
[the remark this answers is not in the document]
lkr-ind: Can you please explain further? Are you saying the whole text is not needed. I would like to keep this text as it helps to understand the overall structure.

apps/dirac/README.md
lkr-ind: Most of the text has been modified in #26. 

I would like to keep the description about the executable name as it will help us/anyone in general in the future.

apps/dirac/Ramses/ramses.py
lkr-ind: I would need more discussion on this one as I did not understand it completely. As you have mentioned that it might be for another PR.

apps/dirac/README.md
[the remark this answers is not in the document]
t-young31: Ah sorry – should clarify that generally the comments only apply to the line immediately above (i.e. L24 here)

apps/dirac/Additional_Input_Data/readme.txt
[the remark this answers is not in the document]
t-young31: Awesome – just checking 😄
If they're public it may be nice to have them on [Zenodo](https://zenodo.org/) or some other cloud storage location that can be `wget`-ed in on demand. Only a suggestion – maybe something for the future @giordano ?

apps/dirac/Ramses/ramses.py
t-young31: I was thinking that the following could be in a base test class (i.e. something like that in #19) 

```python
class BaseTest(rfm.RegressionTest):
    @run_after("something?")
    def check_executable_exists(self):

        if not Path(self.executable).exists():
            raise RuntimeError("Cannot run {self} - executable didn't exist")
```

to prevent a benchmark from queueing only to fail because the executable doesn't exist. Perhaps ReFrame does this already 🤔 ?

lkr-ind: Thanks for clarifying and yes, it is a nice suggestion. I think @giordano might be having more knowledge about this.

I am not sure that Reframe will check about the executable automatically. The test will be queued and will fail because executable is not found. If we insert the above statement, I think, it will check to see if executable exists or not. However, the test may be scheduled and only when it runs it will find about the missing executable (again not entirely sure).

apps/dirac/Additional_Input_Data/readme.txt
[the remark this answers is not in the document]
giordano: Strongly agreed.  If it's ok to have this data publicly accessible, something which doesn't require authentication and can be `wget`-ed straight away would be much better.  We really want the benchmarks to be run reproducibly, not that people have to go around fetching the data manually.  Zenodo looks like a good option: they offer [50 GiB per dataset](https://help.zenodo.org/), I think we're within that limit, right?

apps/grid/grid.py
```diff
@@ -71,8 +71,10 @@ class GridBenchmark_ITT(GridBenchmark):
     @run_after('setup')
     def setup_num_tasks(self):
         self.num_tasks = reduce(mul, list(map(int, self.mpi.split("."))), 1)
-        self.set_var_default('num_cpus_per_task',
-                             self.current_partition.processor.num_cpus)
+        self.set_var_default(
+            'num_cpus_per_task',
+            self.current_partition.processor.num_cpus //
+            min(1, self.current_partition.processor.num_cpus_per_core))
```
giordano, line 77: I also did this change suggested by Lokesh the other day during the meeting we had, related to the fact some systems have `current_partition.processor.num_cpus_per_core > 1` when (presumably) hyperthreading is enabled, and related to the fact the new setting of cosma8 correctly sets `num_cpus` _and_ `num_cpus_per_core`.

PR #29, opened 2022-07-06
Remove intel mpi on csd3

tkoskela:
Since we are getting errors with impi and have to manually ask for openmpi in the spec, might as well default to openmpi


tkoskela:
Ah, great, I branched out from #27 by mistake

giordano:
I'm not so sure anymore this is necessary, both OpenMPI and Intel MPI seems to be working for me now :confused:

tkoskela:
I'm happy to close this without merging

PR #31, opened 2022-07-08
Separate CSD3 into two different systems: Skylake and Icelake

### LINKED ISSUE #28: Separate csd3 icelake and skylake to different systems
They use different versions of the module environment and this can only be specified at system level

giordano:
The two kind of nodes have different operating systems, which causes a few
issues:

* some system libraries have different ABIs (among them, OpenSSL is one
  often problematic)
* the two systems have different versions of the module system (Skylake nodes
  use tmod 3.2, Icelake nodes use tmod 4), which makes it impossible to have
  both of them working properly in ReFrame if they are two partitions of the
  same system.

Fix #28.  CC @lkr-ind.


tkoskela (approved):
Looks OK to me

PR #49, opened 2022-10-17
[archer2] A few changes to the Spack environment

giordano:
See the individual commits for the changes done.  Main troubles I had for building Grid were

* compiling with gcc and using cray-mpich built with craylang (the fact that the modules have the same names is also a bit confusing)
* Cray wrappers for GCC caused lots of confusion for Spack (more specifically when trying to set `SPACK_COMPILER_IMPLICIT_RPATHS`), using the GCC binaries directly solved the issue

On this branch I was able to build `grid@develop%cce` and `grid@develop%gcc` without problems.

### DIFF
--- a/spack-environments/archer2/compute-node/spack.yaml
+++ b/spack-environments/archer2/compute-node/spack.yaml
@@ -193,6 +193,10 @@ spack:
       environment: {}
       extra_rpaths: []
   packages:
+    slurm:
+      externals:
+      - spec: slurm@21.08.8
+        prefix: /usr
     autoconf:
       externals:
       - spec: autoconf@2.69
@@ -233,6 +237,10 @@ spack:
         modules:
         - libfabric/1.11.0.4.71
         - cray-mpich/8.1.15
+    cray-pmi:
+      externals:
+      - spec: cray-pmi@6.1.1
+        prefix: /opt/cray/pe/pmi/6.1.1
     diffutils:
       externals:
       - spec: diffutils@3.6
@@ -287,6 +295,10 @@ spack:
         prefix: /opt/cray/pe/hdf5-parallel/1.12.0.3/gnu/9.1
         modules:
         - cray-hdf5-parallel/1.12.0.3
+    libfabric:
+      externals:
+      - spec: libfabric@1.11.0
+        prefix: /opt/cray/libfabric/1.11.0.4.71
     libtool:
       externals:
       - spec: libtool@2.4.6
--- a/spack-environments/archer2/compute-node/spack.yaml
+++ b/spack-environments/archer2/compute-node/spack.yaml
@@ -219,9 +219,15 @@ spack:
         - cray-libsci/21.04.1.1
     cray-mpich:
       externals:
-      - spec: cray-mpich@8.1.15
+      - spec: cray-mpich@8.1.15%cce
         prefix: /opt/cray/pe/mpich/8.1.15/ofi/crayclang/10.0
         modules:
+        - libfabric/1.11.0.4.71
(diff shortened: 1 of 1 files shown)


spack-environments/archer2/compute-node/spack.yaml
```diff
@@ -132,10 +132,10 @@ spack:
...
       spec: gcc@9.3.0
       paths:
-        cc: cc
-        cxx: CC
-        f77: ftn
-        fc: ftn
+        cc: /opt/cray/pe/gcc/9.3.0/snos/bin/gcc
+        cxx: /opt/cray/pe/gcc/9.3.0/snos/bin/g++
+        f77: /opt/cray/pe/gcc/9.3.0/snos/bin/gfortran
+        fc: /opt/cray/pe/gcc/9.3.0/snos/bin/gfortran
```
aturner-epcc, line 138: @giordano If you use the bare compiler commands in this way (rather than the wrappers) then there is no guarantee things will work as expected. The HPE Cray Programming Environment is all built around the design choice that you never call the compilers directly but, instead, always access them through the wrapper commands. Your approach may work if you set all the library spec (as you have done below) but you may encounter weirdness as they were never designed to be used like this.

spack-environments/archer2/compute-node/spack.yaml
```diff
@@ -283,6 +323,12 @@ spack:
       externals:
       - spec: pkg-config@0.29.2
         prefix: /usr
+    python:
+      externals:
+      - spec: python@2.7.18+bz2+ctypes~dbm+nis+pyexpat+pythoncmd+readline+sqlite3+ssl~tix~tkinter+uuid+zlib
+        prefix: /usr
+      - spec: python@3.6.12+bz2+ctypes~dbm+lzma+nis+pyexpat~pythoncmd+readline+sqlite3+ssl~tix~tkinter+uuid+zlib
+        prefix: /usr
```
aturner-epcc, line 331: You may want to look at the Python 3 environment in the `cray-python` module for a more complete installation

spack-environments/archer2/compute-node/spack.yaml
giordano: The main motivation for going directly to the bare GCC binaries is that yesterday I struggled to build a package using the Cray wrappers for GCC, more specifically Spack wasn't able to detect the compilers implicit rpaths
```console
> spack build-env 'grid@develop%gcc' -- printenv SPACK_COMPILER_IMPLICIT_RPATHS
>
```
because the Cray wrappers were trying, and failing, to lo load some extra modules.  Instead I had no problems with the environment in this PR:
```console
> spack build-env 'grid@develop%gcc' -- printenv SPACK_COMPILER_IMPLICIT_RPATHS
/opt/cray/pe/gcc/11.2.0/snos/lib/gcc/x86_64-suse-linux/11.2.0:/opt/cray/pe/gcc/11.2.0/snos/lib64
>
```
Getting the implicit RPATHs is important because they contain the directory where libstdc++, and trying to run a C++ program built with GCC 11 (like the one I was trying to build yesterday) without the RPATH set correctly would fail because it'd try to use the global libstdc++, which has an older ABI.

If you're specifically worried about Cray MPICH, note that since version 8.1.7 this package comes with its own compiler wrappers, and Spack automatically uses them when available: https://github.com/spack/spack/blob/b21e54dffa7b666462f063ab9d4d51109202b956/var/spack/repos/builtin/packages/cray-mpich/package.py#L32-L33.  This is also the reason why I added to the environment `cray-mpich` v8.1.5.  I verified the right libmpi is being linked in this case, and the application runs fine when launched with `srun`.

PR #50, opened 2022-10-24
[imb] Avoid using Pandas

### LINKED ISSUE #47: Remove imb pandas dependency
The `apps/imb` benchmark has a dependency on pandas in `modules/imb.py`. It is only a single use of `pd.DataFrame` to read in data from the benchmark. Since pandas is not always available by default, e.g. on archer2, we should replace it with a numpy array or something else that is more widely available.

giordano:
While a dataframe is a nicer datastructure for holding the data coming out of the IMB benchmarks, we don't really use dataframes features in any way and this needlessly put a requirement for `pandas` when running the benchmark.  We can easily live with a list of lists, without much effort (actually, with one or two lines less in total)

Fix #47.

### DIFF
--- a/apps/imb/imb.py
+++ b/apps/imb/imb.py
@@ -19,7 +19,7 @@ from modules.imb import read_imb_out
 from modules.reframe_extras import ScalingTest
 from modules.utils import SpackTest
 
-Metric = namedtuple('Metric', ['column', 'function', 'unit', 'label'])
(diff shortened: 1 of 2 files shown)


tkoskela (commented):
Thanks, looks ok to me! Could you put in comments what you expect each column number to correspond to in the output?

giordano:
Done :+1:

PR #51, opened 2022-10-27
Move Intel HPL benchmark to new framework

giordano:
I tried to adapt the old StackHPC Intel HPL benchmark to our new framework using Spack.  I simplified it a bit, it can definitely be improved, but it's a start.  I ran successfully the single-node benchmark on `csd3-skylake` (got `2.01776e+03` as Gflops).

CC @JamieJQuinn


giordano:
This can definitely be improved, but it can be done later, I'd say let's merge now and make changes in follow up PRs.

PR #59, opened 2022-11-29
Require Spack 0.18

giordano:
Also, update the syntax of the concretizer in Spack environments.

PR #60, opened 2022-12-01
Update Spack environment for GitHub Actions

giordano:


giordano:
I was trying to disentagle the update of the environment from #59, but both are needed together to make CI work, so can't really separate the two changes.  Going to close this PR.

PR #85, opened 2023-02-02
Add custom repo for Spack packages

### LINKED ISSUE #83: Add directory for Spack packages
So that packages that aren't builtins to the Spack trunk can be used

giordano:
I did a quick test locally and it seemed to work as expected, but need more thorough testing.  And probably a quick explanation in the docs of where to put the packages, but basically you need to place the `package.py` in a subdirectory inside `repo/packages/`, for example `repo/packages/libfoo/package.py` for `libfoo`.

Fix #83.


dcaseGH:
Mose,
I've had a quick look at this, and I'll explain what I did. 
- I checked out your branch
- Then I set up my environment:
```
export SPACK_ROOT="/home/mox-dcase/spack"
source "${SPACK_ROOT}/share/spack/setup-env.sh"
. "/home/mox-dcase/miniconda3/etc/profile.d/conda.sh"
conda activate hpc-tests
export RFM_CONFIG_FILE="/home/mox-dcase/SolverBenchmarks/excalibur-tests/reframe_config.py"
export "PATH=/home/mox-dcase/SolverBenchmarks/excalibur-tests/reframe/bin:$PATH"
```
- I ran `reframe -c examples/sombrero -r` which was fine
- Then, as a test of the new approach, I moved the spack package for sombrero from the builtins to spack-environments/repo/packages/ and to be sure I was testing things I changed the name of the class and the directory to be `sombrero2`
- Initially this fails as spack can't find the repo. But with `spack repo add spack-environments/repo` the reframe tests pass
- P.S. I am not suggesting we move things from builtins generally - I just wanted a test which should work. I will put new things in the new repo directory going forwards

Thanks for doing this: 
- please let me know if I've done what you hoped.
- we could also have some standard way of setting things up to run reframe? I will add `spack repo add spack-environments/repo` to my environment set-up - I will look at the documentation that we provide to see if this would all be clear to other users
- there are still mix-ups that easily occur for me, for e.g. the RFM_CONFIG_FILE that I used was from the directory above my reframe installation, and differed slightly from the branch I was testing (although I don't think it affects the above). I should make sure that I'm running reframe exactly as intended, as reproducibility is half of the aim here ;)

giordano:
@dcaseGH I _think_ the problem was that the path of the repo was a bit ill-defined, as it'd depend on the current working directory.  I'm now using the [Spack-specific variable](https://spack.readthedocs.io/en/latest/configuration.html#spack-specific-variables) `$env` to set the path of the repo, this should make this more robust.  Can you please give the new version a go?

giordano:
I've been doing some tests using this PR + #97 and I think this is now working as I expect.  @dcaseGH let me know if this works for you as well now.

dcaseGH:
@giordano Do you mind clarifying this for me? What should I be setting? The `env` variable? Is this set within your branch, or in your copy of spack?
Previously I was using `spack repo add spack-environments/repo`.
Thanks

PR #90, opened 2023-02-08
Update documentation

ilectra:
Especially installation and running instructions in the main README.


README.md
```diff
@@ -7,37 +7,39 @@ These benchmarks are based on a similar project by
 
 _**Note**: at the moment the ExCALIBUR benchmarks are a work-in-progress._
 
+## Before you start
+
+If you are using spack for other projects, you might have to clear the cache if it exists:
+`rm -rf ~/.spack/cache`
```
giordano, line 13: `spack clean -m` is a better way to clear the cache, instead of suggesting to `rm -rf` directories which may not even exist (for example on Acher2 and other Cray system you typically don't have anything related to spack in your home directory since it can't be mounted at all by compute nodes).

Also, I don't think the "using spack for other projects" description is accurate, you may have troubles when upgrading spack after some time, especially in upgrading past a breaking version (e.g. you were on 0.19 and now checked out a recent `development` branch which is on 0.20), they occasionally change the names of some packages, and if they stick in the cache then the resolve gets very disappointed.  And yes, this is a bit annoying.  Using Spack for other projects is irrelevant, as far as I know.

README.md
```diff
@@ -7,37 +7,39 @@ These benchmarks are based on a similar project by
...
 
 ### Spack
 
-_**Note**: we require Spack 0.18.  In some HPC facilities there may be already a
-central Spack installation available.  In principle you should be able to use a
-system installation of Spack (you only need to have `spack` in the `PATH`), but
-you need to make sure the version is recent enough to install the required
-packages and understand the provided environments.  Instructions below show you
-how to install Spack locally._
+_**Note**: In some HPC facilities there may be already a central Spack installation available.
```
giordano, line 19: Uhm, we do require at least Spack 0.18: #59.  Why was that removed?  All what we do should also be compatible with 0.19 and 0.20, that may be made clearer.

README.md
```diff
@@ -7,37 +7,39 @@ These benchmarks are based on a similar project by
...
 [Spack](https://spack.io/) is a package manager specifically designed for HPC
-facilities.  Follow the [official
+facilities.  The [official
 instructions](https://spack.readthedocs.io/en/latest/getting_started.html) to
-install the latest version of Spack.
+install the latest version of Spack are summarised here for convenience.
+- git clone spack:
+`git clone -c feature.manyFiles=true https://github.com/spack/spack.git`
+- run spack setup script: `source ./spack/share/spack/setup-env.sh`
+- check spack is in `$PATH`, for example `spack --version`
```
giordano, line 31: I'm not against putting the instructions here, but for context, the reason why I didn't copy them was not to duplicate information which is already in the official manual and which has a (admittedly very low) risk to become outdated.

ilectra:
I'm not done yet, please wait! :smile:

README.md
ilectra: It's safer to install the latest version anyway, so they don't have to worry about which one is installed on their system.

ilectra:
I fixed the conflicts with main, so I can merge this now. I haven't messed with the Reframe part much, anticipating the pip-installable changes. Please review and merge if happy.

pineapple-cat (approved):
Looks fairly clear to me as well, and including a summary of the Spack install instructions is a nice touch.

README.md
```diff
@@ -59,7 +63,7 @@ install this package.  Note that ReFrame requires Python 3.6: in your HPC system
 you may need to load a specific module to have this version of Python available.
 
 We provide a ReFrame configuration file with the settings of some systems that
```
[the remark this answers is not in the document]
ilectra: Yes.

PR #91, opened 2023-02-13
Make project pip installable

### LINKED ISSUE #17: Make excalibur-tests pip installable?
Currently the setup is a little involved and could perhaps be made simpler by making a setup script, either a setup.py in the root directory to make the whole repo pip installable or maybe a shell script that would:

- install python dependencies
- ~modify your bash_profile with the SPACK_ROOT~ not needed: https://github.com/eth-cscs/reframe/pull/2123
- copy the correct reframe config to ~/.reframe/settings.py 
- install all the extra [python modules](https://github.com/ukri-excalibur/excalibur-tests/tree/main/modules), so there wouldn't need to be any appending to sys.path (which relies on a consistent directory structure)
- prompt user to install spack, if it's not already present

tkoskela:
Closes #17

Adds `pyproject.toml`. You can install the package using `pip install .` to have pip install reframe and have the python modules in `modules.py` importable. As I've added in README.md, It is recommended to install in a virtual environment, using `python3 -m venv <path/to/my/env> && source <path/to/my/env>/bin/activate && pip install .`. To install in development mode that allows editing installed files, use `pip install . -e`. To install in a custom directory (normally goes to myenv/lib/pythonx.x/site-packages) use `pip install --target <custom_dir> .`. I had to use a custom directory on GHA and Archer2 where the install path becomes too long for spack to handle.

Based on David's feedback, I've moved all the files that are installed with the package to a subdirectory, currently called `excalibur_tests`. That makes it clearer what is part of the package and what isn't, both to humans and to `pip`. I might be inclined to change the name because

1) it lengthens the install path
2) it is confusing with respect to the name of the package `excalibur-tests` (python doesn't allow a `-` in the import statement)

The directory name will be used when importing modules from the package, ie. `import excalibur_tests.modules.utils`. I'm open to suggestions what to call the subdirectory, some of my ideas are: `benchmarks`, `tests` and `framework`

- [x] Test on remote system(s)
- [x] Update README
- [x] Update tests
- [x] Decide whether to rename package directory


pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+]
+description = "Framework for building and running HPC benchmarks"
+readme = "README.md"
+requires-python = ">=3.6"
+keywords = ["benchmark", "hpc"]
+license = {text = "Apache-2.0"}
+classifiers = [
+    "Programming Language :: Python :: 3",
+]
+dependencies = ["reframe-hpc==3.11.0"]
```
giordano, line 22: ReFrame 3.12 is also ok.  It'd be something like `>= 3.11.0, < 4.0.0` (I don't know the syntax, I'm making it up)

pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[tool.setuptools.packages.find]
+where = ["modules"]
+include = ["*"]
+exclude = []
+namespaces = false
```
giordano, line 28:
```suggestion
namespaces = false

```

giordano:
Uhm, this just occurred to me: if we install this via pip, _where_ is it installed?  Is it the user's home directory?  If so, can this installation directory be chosen by the user?  On Cray systems like ARCHER2 the home directory is ***not*** mounted on compute nodes, and for example we have to work around this limitation in https://github.com/ukri-excalibur/excalibur-tests/blob/6b7d83f86da91ff4b07b307ae2d45f519d3a1a2b/reframe_config.py#L37-L44

tkoskela:
On my local machine I use `venv` and pip installs inside the directory of the environment that can be anywhere.

On Archer2 specifically they have [recommendations for installing Python packages](https://docs.archer2.ac.uk/user-guide/python/) by either setting `PYTHONUSERBASE` to point at your work directory, or using virtual environments. I'll add documentation to this PR on setting up a virtual environment.

giordano (commented):
Other question: how is this supposed to work for development, i.e. when I do _not_ want to install this via pip?  The `sys.path.append` was a bad hack, and I'm glad that's gone, but what am I supposed to do if I want to edit the `SpackTest` class locally?

tkoskela:
> Other question: how is this supposed to work for development, i.e. when I do _not_ want to install this via pip? The `sys.path.append` was a bad hack, and I'm glad that's gone, but what am I supposed to do if I want to edit the `SpackTest` class locally?

For [development](https://setuptools.pypa.io/en/latest/userguide/development_mode.html), you can use `pip install -e`. That will install the package by linking to the files in your local directory

giordano:
> For [development](https://setuptools.pypa.io/en/latest/userguide/development_mode.html), you can use `pip install -e`. That will install the package by linking to the files in your local directory

Would be good to document this?

tkoskela:
> > For [development](https://setuptools.pypa.io/en/latest/userguide/development_mode.html), you can use `pip install -e`. That will install the package by linking to the files in your local directory
> 
> Would be good to document this?

Yep, will add it to README of this PR

apps/dirac/Sphng/sphng.py
```diff
@@ -1,5 +1,3 @@
-import os
```
giordano, line 1: `os` is used here

apps/wrf/wrf.py
```diff
@@ -1,15 +1,11 @@
 # TODO: Add docs
 # TODO: Currently must set WRF_DIR in environment's variables (to parent of /RUN directory) - would be better to use a module?
 
-import os
```
giordano, line 4: `os` is used here.

giordano (commented):
You can't remove `os` from everywhere, sometimes it's used for something else.

modules/utils.py
```diff
@@ -226,7 +226,7 @@ def identify_build_environment(current_partition):
     else:
         system, partition = current_partition.fullname.split(':')
         cp_dir = os.path.realpath(
-            os.path.join(os.path.dirname(__file__), '..', 'spack-environments',
+            os.path.join(os.path.dirname(__file__), '..', 'envs',
```
dpshelio, line 229: `pathlib` is more readable than `os.path`. So I'd change this to:
```
from pathlib import Path

cp_dir = Path(__file__).absolute().parent.parent / 'envs').
```

Then you can do on 232:
```
env_dir = cp_dir / partition
```

pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+    "Programming Language :: Python :: 3",
+]
+dependencies = [
+    "reframe-hpc >= 3.11.0, < 4.0.0",
+    "matplotlib >= 3.0.0",
+]
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["modules","apps","examples","envs","perflogs"]
```
dpshelio, line 32: I would suggest putting all these directories under a directory name, that then is the same as the package name. So people can do `import excalibur_test`  or internally makes it easier (e.g, examples) to load it.
It will also make that line easier by including that directory.

apps/wrf/wrf.py
```diff
@@ -4,12 +4,9 @@
 import os
 import re
 import string
-import sys
 import urllib
 import reframe as rfm
 import reframe.utility.sanity as sn
-
-sys.path.append(os.path.join(os.path.dirname(__file__), '..', '..'))
```
dpshelio, line 12: this is not needed and you could use relative imports as
```
from ...modules.utils import SpackTest
```
But it's probably cleaner using absolute imports instead.

dpshelio (commented):
My main suggestion would be to create a common directory and remove the `sys.path` from the library by using relative or absolute imports.

examples/sombrero/sombrero.py
```diff
@@ -3,14 +3,9 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
-import os.path as path
-import sys
 import reframe as rfm
 import reframe.utility.sanity as sn
 
-# Add top-level directory to `sys.path` so we can easily import extra modules
-# from any directory.
-sys.path.append(path.join(path.dirname(__file__), '..', '..'))
```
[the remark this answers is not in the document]
tkoskela: This PR is removing the `sys.path.append`s everywhere :)

tkoskela:
> On a third thought: maybe this change should be done in a separate PR.

I approve of this idea. I think I should be able to revert 0adc05f from this PR. I added it because tests were failing on GHA but fec481c shortened the install path so it's probably fine :shrug:

tkoskela:
I reverted the renaming commits, merged #99, and moved the package modules into `excalibur_tests` again. Nothing seems to be broken. We should be able to rename directories in a later PR.

tkoskela:
Closed in favour of #102

PR #93, opened 2023-02-20
[spack] Move `intall_tree` option back to common file

giordano:
While looking into https://github.com/ukri-excalibur/excalibur-tests/pull/85#issuecomment-1429519523 I found the [Spack-specific variable](https://spack.readthedocs.io/en/latest/configuration.html#spack-specific-variables) `$env`, which allows us to reference the path to the active environment in the common file `common.yaml`, so that we can have again the `install_tree` setting in the common file.  This more or less reverts  #75, but now we have better solution.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ Here are the steps to create a Spack environment for a new system:
 * set the
   [`install_tree`](https://spack.readthedocs.io/en/latest/config_yaml.html#install-tree):
   ```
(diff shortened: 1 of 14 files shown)


tkoskela (approved):
I like having the `install_tree` option in one place!

giordano:
Yeah, that was also the case between #59 and #75, but I wasn't using `$env`, so that wasn't working very well, but this PR should have fixed it.

PR #94, opened 2023-02-20
Update README.md

tkoskela:
Add documentation on how to run spack and reframe locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,29 @@ the "generic" system, a new empty Spack environment will be automatically create
 different Spack environment by setting the environment variable `EXCALIBUR_SPACK_ENV`
 described above.
 
+### Local usage
+
+You can use the framework to run tests on a local system without a scheduler. After spack has
+been installed following the instructions given above, it needs to be told about the software
+installed on the system. On supported HPC systems this is handled by the environment. By default an
+empty spack environment will be created in `spack_environments/generic/default` when you run
+ReFrame for the first time .To find locally installed
+compilers run `spack compiler find`. To find locally installed packages run `spack external find`. This
+does not necessarily find everything and some packages may have to be added manually, in particular it is 
+worth making sure that a local MPI
+installation is added to the environment. If no MPI package has been added in `~/.spack/packages.yaml`,
+add
+```
+  packages:
+    openmpi:
+      externals:
+      - spec: <your installed MPI, e.g. openmpi@4.1.2>
+        prefix: <path to MPI installation, e.g. /usr>
+```
+To either `spack_environments/generic/default/spack.yaml` or `~/.spack/packages.yaml`. The name,
+version, and prefix may vary depending on your local operating system. The examples given above
+are the defaults on Ubuntu 22.04.
+
 ## Contributing new systems or benchmarks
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,6 +161,29 @@ the "generic" system, a new empty Spack environment will be automatically create
 different Spack environment by setting the environment variable `EXCALIBUR_SPACK_ENV`
 described above.
 
+### Local usage
+
+You can use the framework to run tests on a local system without a scheduler. After spack has
+been installed following the instructions given above, it needs to be told about the software
+installed on the system. On supported HPC systems this is handled by the spack environment. By default an
+empty spack environment will be created in `spack_environments/generic/default` when you run
```
giordano, line 169:
```suggestion
empty spack environment will be created in `spack-environments/generic/default` when you run
```
Side note, we should probably shorten the name of the `spack-environments` directory.

README.md
```diff
@@ -161,6 +161,29 @@ the "generic" system, a new empty Spack environment will be automatically create
...
+ReFrame for the first time .To find locally installed
```
giordano, line 170:
```suggestion
ReFrame for the first time. To find locally installed
```

README.md
```diff
@@ -161,6 +161,29 @@ the "generic" system, a new empty Spack environment will be automatically create
...
+compilers run `spack compiler find`. To find locally installed packages run `spack external find`. This
+does not necessarily find everything and some packages may have to be added manually, in particular it is 
+worth making sure that a local MPI
+installation is added to the environment. If no MPI package has been added in `~/.spack/packages.yaml`,
```
giordano, line 174: Don't reference `~/.spack`, it may not even exist: https://github.com/ukri-excalibur/excalibur-tests/pull/90#discussion_r1100314544

tkoskela:
Some of this stuff on setting up the environment is already in `contributing.md` but it may not be an obvious place to look. Should I just reference it here instead of saying the same things twice?

giordano:
> Some of this stuff on setting up the environment is already in contributing.md but it may not be an obvious place to look.

Yeah, I feel like `CONTRIBUTING.md` might be a better place.  Maybe add a reference in the `README.md` about this kind of information being available there?

README.md
```diff
@@ -164,28 +164,14 @@ described above.
...
-```yaml
-  packages:
-    <your installed MPI implementation, e.g. openmpi>:
-      externals:
-      - spec: <your installed MPI implementation@version, e.g. openmpi@4.1.2>
-        prefix: <path to MPI installation, e.g. /usr>
-```
-To either `spack-environments/generic/default/spack.yaml` or `packages.yaml`. The name,
-version, and prefix may vary depending on your local operating system. The examples given above
-are the defaults on Ubuntu 22.04.
```
ilectra: Everything above beginning with "This does not necessarily find everything..." should be either here or in the `CONTRIBUTING.md`. Probably the latter.

tkoskela: I've removed that from here and refer to `CONTRIBUTING.md`

tkoskela:
"Unsupported systems" still sounds a bit unfriendly but I couldn't come up with what a better name

ilectra (approved):

PR #96, opened 2023-02-22
[Isambard-a64fx] Force compilation for A64FX in Spack

giordano:
Spack currently misdetects Isambard A64FX as a generci Aarch64 target.  This change forces the compilation for A64FX.

### DIFF
--- a/spack-environments/isambard-a64fx/compute-node/spack.yaml
+++ b/spack-environments/isambard-a64fx/compute-node/spack.yaml
@@ -8,6 +8,9 @@ spack:
   view: true
   include:
   - ../../common.yaml
+  concretizer:
+    targets:
+      host_compatible: False
   compilers:
   - compiler:
       spec: cce@10.0.3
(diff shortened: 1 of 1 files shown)


spack-environments/isambard-a64fx/compute-node/spack.yaml
```diff
@@ -8,6 +8,9 @@ spack:
   view: true
   include:
   - ../../common.yaml
+  concretizer:
+    targets:
+      host_compatible: False
```
giordano, line 13: For the record, this is necessary only because Spack misdetects the A64FX CPU on Isambard as generic Aarch64, because it uses an old Cray programming environment, see

* https://github.com/archspec/archspec-json/pull/59#issuecomment-1339279316
* https://github.com/archspec/archspec-json/issues/23#issuecomment-1340882940
* https://github.com/archspec/archspec-json/issues/23#issuecomment-1415799814

PR #95, opened 2023-02-21
[isambard-cascadelake] Various improvements to Spack environment

giordano:
@kaanolgu Can you please try out this environment?  It worked for me to install `libxml2%gcc@9.2.0`

PR #97, opened 2023-02-22
ReFrame Package for Babelstream

kaanolgu:
Gives an error for the `+opencl` version on Isambard `cascadelake` and `broadwell` architectures while using the environment from `excalibur-tests/spack-environments`

```
==> Installing perl-5.36.0-2onqezf52p53b24p6hkwszxr7niup7dw
==> No binary for perl-5.36.0-2onqezf52p53b24p6hkwszxr7niup7dw found: installing from source
==> Error: IndexError: list index out of range

/lustre/home/br-kolgu/spack/var/spack/repos/builtin/packages/perl/package.py:425, in setup_build_environment:
        422        # This is how we tell perl the locations of bzip and zlib.
        423        env.set("BUILD_BZIP2", 0)
        424        env.set("BZIP2_INCLUDE", spec["bzip2"].prefix.include)
  >>    425        env.set("BZIP2_LIB", spec["bzip2"].libs.directories[0])
        426        env.set("BUILD_ZLIB", 0)
        427        env.set("ZLIB_INCLUDE", spec["zlib"].prefix.include)
        428        env.set("ZLIB_LIB", spec["zlib"].libs.directories[0])


==> Warning: Skipping build of libxcrypt-4.4.33-s5pffjcgnux72cismwavlcjniakyegga since perl-5.36.0-2onqezf52p53b24p6hkwszxr7niup7dw failed
==> Warning: Skipping build of python-3.10.8-ooo4lxxdlutn56ihasc76zjvdimrzovq since libxcrypt-4.4.33-s5pffjcgnux72cismwavlcjniakyegga failed
==> Warning: Skipping build of ninja-1.11.1-eorn7rovlt2b3qvib7oulbwili2v3xcj since python-3.10.8-ooo4lxxdlutn56ihasc76zjvdimrzovq failed
==> Warning: Skipping build of llvm-10.0.1-zv7o3m4plbxxy56n7fr7dbassswihe35 since ninja-1.11.1-eorn7rovlt2b3qvib7oulbwili2v3xcj failed
==> Warning: Skipping build of pocl-1.5-vz5qvimab3dev3a6qshf4u57uiacafmt since llvm-10.0.1-zv7o3m4plbxxy56n7fr7dbassswihe35 failed
==> Warning: Skipping build of babelstream-main-ta2zrur3f4pjocf6sc6ozn3byhkx43p6 since pocl-1.5-vz5qvimab3dev3a6qshf4u57uiacafmt failed
==> Warning: Skipping build of texinfo-7.0-jt7o7y4gdtwgafzjba4oe6fby74zaymh since perl-5.36.0-2onqezf52p53b24p6hkwszxr7niup7dw failed
==> Warning: Skipping build of gcc-9.2.0-srd2b46dflcceslfap6zediyt3gd3zp3 since texinfo-7.0-jt7o7y4gdtwgafzjba4oe6fby74zaymh failed
==> Warning: Skipping build of binutils-2.40-7t7umksnqcj2y44f5nwiwumcb7tmvumb since texinfo-7.0-jt7o7y4gdtwgafzjba4oe6fby74zaymh failed
==> Warning: Skipping build of perl-data-dumper-2.173-or5mfn64xjtq4waxm54gc3ieoibmnalm since perl-5.36.0-2onqezf52p53b24p6hkwszxr7niup7dw failed
==> Warning: Skipping build of openssl-1.1.1s-ph24dhbil7ojmbhga2uyck2reelc2mdi since perl-5.36.0-2onqezf52p53b24p6hkwszxr7niup7dw failed
==> Warning: Skipping build of cmake-3.23.2-q6yybin3rel27zbudr6dahndwo7qkdrz since openssl-1.1.1s-ph24dhbil7ojmbhga2uyck2reelc2mdi failed
==> Error: babelstream-main-ta2zrur3f4pjocf6sc6ozn3byhkx43p6: Package was not installed
==> Error: Installation request failed.  Refer to reported errors for failing package(s).
```

This is the result of concretize : 
```
$ spack -e ~/excalibur-tests/spack-environments/isambard-cascadelake/compute-node/ spec -l babelstream%gcc@9.2.0 +opencl
Input spec
--------------------------------
babelstream%gcc@9.2.0+opencl

Concretized
--------------------------------
3ldokiu  babelstream@main%gcc@9.2.0~acc~cuda~ipo~kokkos~omp+opencl~raja~rocm~stddata~stdindices~stdranges~sycl~sycl2020~tbb~thrust build_system=cmake build_type=RelWithDebInfo compiler=none cpu_arch=none dir=none flags=none mem=DEFAULT target=none arch=cray-rhel8-broadwell
to2hagx      ^cmake@3.11.4%gcc@9.2.0~doc+ncurses+ownlibs~qt build_system=generic build_type=Release patches=1c54004,93488ed,dd3a40d arch=cray-rhel8-broadwell
t3emi7a      ^gcc@9.2.0%gcc@9.2.0~binutils+bootstrap~graphite~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages=c,c++,fortran patches=49341f7,98a9c96,dc1ca24 arch=cray-rhel8-broadwell
2hi3yhw          ^diffutils@3.6%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
suyaagj          ^gawk@5.1.1%gcc@9.2.0~nls build_system=autotools arch=cray-rhel8-broadwell
nafck73              ^libsigsegv@2.13%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
yrsby5g              ^readline@8.2%gcc@9.2.0 build_system=autotools patches=bbf97f1 arch=cray-rhel8-broadwell
fty63tk          ^gmp@6.2.1%gcc@9.2.0+cxx build_system=autotools libs=shared,static arch=cray-rhel8-broadwell
w2m3xbj              ^autoconf@2.69%gcc@9.2.0 build_system=autotools patches=7793209 arch=cray-rhel8-broadwell
il4ujg6              ^automake@1.16.1%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
ox3z3g4              ^m4@1.4.18%gcc@9.2.0+sigsegv build_system=autotools patches=3877ab5,fc9b616 arch=cray-rhel8-broadwell
lsp62nw          ^libtool@2.4.6%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
lmopn6h          ^mpc@1.1.0%gcc@9.2.0 build_system=autotools libs=shared,static arch=cray-rhel8-broadwell
vauckwc          ^mpfr@3.1.6%gcc@9.2.0 build_system=autotools libs=shared,static patches=7a6dd71 arch=cray-rhel8-broadwell
3a3jpw5          ^texinfo@7.0%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
r6n3ke5              ^gettext@0.19.8.1%gcc@9.2.0+bzip2+curses+git~libunistring+libxml2+tar+xz build_system=autotools patches=9acdb4e arch=cray-rhel8-broadwell
562jkmd              ^ncurses@6.4%gcc@9.2.0~symlinks+termlib abi=none build_system=autotools arch=cray-rhel8-broadwell
rnvbzgx              ^perl@5.36.0%gcc@9.2.0+cpanm+open+shared+threads build_system=generic arch=cray-rhel8-broadwell
2hxwrio                  ^berkeley-db@18.1.40%gcc@9.2.0+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc arch=cray-rhel8-broadwell
qas5asi                  ^bzip2@1.0.6%gcc@9.2.0~debug~pic+shared build_system=generic arch=cray-rhel8-broadwell
vdyemok                  ^gdbm@1.23%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
vhlozyq          ^zlib@1.2.13%gcc@9.2.0+optimize+pic+shared build_system=makefile arch=cray-rhel8-broadwell
dw7sywx      ^pocl@1.5%gcc@9.2.0~distro~icd~ipo build_system=cmake build_type=RelWithDebInfo arch=cray-rhel8-broadwell
hl4mcwv          ^hwloc@2.9.0%gcc@9.2.0~cairo~cuda~gl~libudev+libxml2~netloc~nvml~oneapi-level-zero~opencl+pci~rocm build_system=autotools libs=shared,static arch=cray-rhel8-broadwell
j4zkygo              ^libpciaccess@0.16%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
uyhxxut                  ^util-macros@1.19.3%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
qc25kxy              ^libxml2@2.10.3%gcc@9.2.0~python build_system=autotools arch=cray-rhel8-broadwell
h7xbdr7                  ^libiconv@1.17%gcc@9.2.0 build_system=autotools libs=shared,static arch=cray-rhel8-broadwell
5253twk                  ^xz@5.2.4%gcc@9.2.0~pic build_system=autotools libs=shared,static arch=cray-rhel8-broadwell
p5gr4qu          ^llvm@10.0.1%gcc@9.2.0+clang+compiler-rt~cuda+gold+internal_unwind~ipo+libcxx~link_llvm_dylib+lld+lldb+llvm_dylib~mlir~omp_debug~omp_tsan+polly~python~split_dwarf~z3 build_system=cmake build_type=Release patches=4728466,50bfc4e,514926d,6811ad4,7b286b1,f920173 targets=none version_suffix=none arch=cray-rhel8-broadwell
js6yyyp              ^binutils@2.40%gcc@9.2.0~gas+gold~gprofng~headers~interwork+ld~libiberty~lto~nls~pgo+plugins build_system=autotools libs=shared,static arch=cray-rhel8-broadwell
6arxdbq              ^libedit@3.1-20210216%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
si7uwhg              ^ninja@1.11.1%gcc@9.2.0+re2c build_system=generic arch=cray-rhel8-broadwell
hnj2ode                  ^re2c@2.2%gcc@9.2.0 build_system=generic arch=cray-rhel8-broadwell
k5stws6              ^perl-data-dumper@2.173%gcc@9.2.0 build_system=perl arch=cray-rhel8-broadwell
wgtzw4l              ^python@3.10.8%gcc@9.2.0+bz2+crypt+ctypes+dbm~debug+libxml2+lzma~nis~optimizations+pic+pyexpat+pythoncmd+readline+shared+sqlite3+ssl~tkinter+uuid+zlib build_system=generic patches=0d98e93,7d40923,ebdca64,f2fd060 arch=cray-rhel8-broadwell
avldhv5                  ^expat@2.5.0%gcc@9.2.0+libbsd build_system=autotools arch=cray-rhel8-broadwell
6h5sxtg                      ^libbsd@0.11.7%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
iwosnyh                          ^libmd@1.0.4%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
a6un3hf                  ^libffi@3.4.4%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
4iazcvw                  ^libxcrypt@4.4.33%gcc@9.2.0~obsolete_api build_system=autotools arch=cray-rhel8-broadwell
rad32hc                  ^openssl@1.1.1s%gcc@9.2.0~docs~shared build_system=generic certs=mozilla arch=cray-rhel8-broadwell
54misz3                      ^ca-certificates-mozilla@2023-01-10%gcc@9.2.0 build_system=generic arch=cray-rhel8-broadwell
grvov5t                  ^sqlite@3.40.1%gcc@9.2.0+column_metadata+dynamic_extensions+fts~functions+rtree build_system=autotools arch=cray-rhel8-broadwell
tafvobi                  ^util-linux-uuid@2.38.1%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
ggoqr5l          ^pkgconf@1.4.2%gcc@9.2.0 build_system=autotools arch=cray-rhel8-broadwell
```


giordano:
All problems with compilation of LLVM on the cascadelake partition should have been fixed by #95 (reference: https://github.com/ukri-excalibur/excalibur-tests/pull/95#issuecomment-1440415056).

This PR depends on #85: we'll put there all our custom Spack packages, before upstreaming them.

apps/babelstream/package.py
```diff
@@ -128,8 +128,9 @@ class Babelstream(CMakePackage,CudaPackage,ROCmPackage):
     # depends_on("raja", when="+raja")
 
     #OpenCL Dependency
-    depends_on("pocl@1.5", when="+opencl")
-    depends_on("gcc@9.2.0", when="+opencl")
+    depends_on("pocl@1.5", when="+ocl")
+    depends_on("gcc@9.2.0", when="+ocl")
+    depends_on("cmake@3.23.2", when="+ocl")
```
kaanolgu, line 133: We need this change because it picks up the `cmake@3.11` from the environment and it does not satisfy the conditions for this build ( needs to be at least 3.14) so either I provide for all options to use `3.23.2` on `package.py` or better option change it directly from environment if other builds won't require this specific version ? @giordano

giordano: If you need `3.14` you can specify exactly that version, instead of the overly restrictive `3.23.2`: if a system happens to have CMake 3.14 can use the global installation, otherwise a newer CMake will be automatically compiled.

Also, I presume CMake is a build-only dependency, so you should add `type="build"`.

giordano: Also, why is CMake a conditional dependency?  Don't you always need it?

apps/babelstream/package.py
```diff
@@ -0,0 +1,363 @@
...
+    depends_on("gcc@12.1.0", when="+stddata")
+
+    depends_on("cmake@3.23.2", when="+stdindices")
+    depends_on("gcc@12.1.0", when="+stdindices")
+    
+    depends_on("cmake@3.23.2", when="+stdranges")
+    depends_on("gcc@12.1.0", when="+stdranges")
+    
+    #OMP Dependency
+    depends_on("openmpi", when="+omp")
```
giordano, line 125: `openmpi` isn't OpenMP, am I missing something else?

apps/babelstream/package.py
kaanolgu: I added it as a dependency because the environment in [here](https://github.com/ukri-excalibur/excalibur-tests/blob/c2b80844eb7096842c595dac07926011659f72a1/spack-environments/isambard-cascadelake/compute-node/spack.yaml#L100-L103) has cmake 3.11.4 so I removed that in my own local version and added `depends_on` but as you said if we remove it from environment it will pick up the latest one so we do not need to mention the version

apps/babelstream/package.py
kaanolgu: It was from the original PR code from [here](https://github.com/spack/spack/pull/32680/files#diff-017e0b85946d9e442915b5b2435f385c86545eeb3fa11c17f7b083fff9a64dafR73) but OpenMP and openMPI are different and actually there is no package for `openMP`. so we could delete this line it is irrelevant and actually we don't need any extra dependency for openMP. Thanks for pointing this out

apps/babelstream/package.py
```diff
@@ -0,0 +1,363 @@
...
+
+    #Thrust Conflict
+    conflicts("~cuda", when="+thrust", msg="Thrust requires +cuda variant")
+
+    #TBB Dependency
+    depends_on("intel-oneapi-tbb", when="+tbb")
+
+    #depends_on("cmake@3.8", when="+stdindices")
+    depends_on("cmake@3.23.2", when="+stddata")
+    depends_on("gcc@12.1.0", when="+stddata")
```
kaanolgu, line 116: Known issue with `gcc@9.2.0` also shared [here](https://github.com/compiler-explorer/compiler-explorer/issues/2075) it won't work with gcc@9.2.0 gives the following error. So we need  `gcc@12.1.0` on the environment too because adding like a dependency this way won't force it. So priority is always given to environment settings. @giordano 

` >> 90    /cm/local/apps/gcc/9.2.0/include/c++/9.2.0/pstl/parallel_backend_tbb.h:19:10: fatal error: tbb/blocked_range.h: No such file or directory
    19| #include <tbb/blocked_range.h>
       compilation terminated.`

giordano: If you have problems with compilers, instead of using [`depends_on()`](https://spack.readthedocs.io/en/latest/packaging_guide.html#depends-on) with that compiler (which would make it complicated to build with *other* compilers), you should use [`conflicts()`](https://spack.readthedocs.io/en/latest/packaging_guide.html#conflicts).

apps/babelstream/package.py
```diff
@@ -0,0 +1,337 @@
...
+    
+    #Kokkos Dependency
+    depends_on("kokkos@3.7.1", when="+kokkos")
+
+    #OpenCL Dependency
+    depends_on("pocl@1.5", when="+ocl")
+
+
+
+    depends_on("cmake@3.14.0:3.23", type="build")
```
giordano, line 107: No need to restrict future versions
```suggestion
    depends_on("cmake@3.14.0:", type="build")
```

apps/babelstream/package.py
```diff
@@ -0,0 +1,337 @@
...
+    #   spack install babelstream +thrust +cuda cuda_arch=70
+    #
+    # if user enters  spack install babelstream +omp : it will build for CPU
+    # if user enters spack install babelstream +omp cuda_arch=70  it will build it for NVIDIA gpu for sm_70  automatically
+    # if user enters spack install babelstream +omp amdgpu_target=gfx701  it will build it for AMD gpu gfx701 automatically
+    #
+    #
+    # TODO: Test installation after it completes and verify
+    # ----------------------------------------------------------------------------
+import os
```
giordano, line 24: This seems unused?

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+    variant("ocl", default=False, description="Enable OpenCL support")
+    variant("tbb", default=False, description="Enable TBB support")
+    variant("acc", default=False, description="Enable OpenACC support") 
+    variant("thrust", default=False, description="Enable THRUST support")
+    variant("raja", default=False, description="Enable RAJA support") 
+    variant("stddata", default=False, description="Enable STD-data support")
+    variant("stdindices", default=False, description="Enable STD-indices support")
+    variant("stdranges", default=False, description="Enable STD-ranges support")
+    
+    variant("dir", values=str, default="none", description="Enable Directory support")
```
tomdeakin, line 40:
```suggestion
    # Some models need to have the programming model abstraction downloaded -
    # this variant enables a path to be provided.
    variant("dir", values=str, default="none", description="Enable Directory support")
```

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+    conflicts("~cuda", when="+thrust", msg="Thrust requires +cuda variant")
+
+    #TBB Dependency
+    depends_on("intel-oneapi-tbb", when="+tbb")
+    
+    #Kokkos Dependency
+    depends_on("kokkos@3.7.1", when="+kokkos")
+
+    #OpenCL Dependency
+    depends_on("pocl@1.5", when="+ocl")
```
tomdeakin, line 72: For OpenCL, we want to build against multiple OpenCL implementations:
- CUDA (provided by spack package `cuda`)
- AMD (I think provided by spack package `rocm-opencl`)
- Intel (?)
- and POCL

So this is really like a backend module, so we might want to say e.g., `+ocl +cuda`

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+  
+    # this flag could be used in all required languages
+    variant("flags", values=str, default="none", description="Additional CXX flags to be provided")
+
+    # SYCL variant
+    variant("compiler", values=str, default="none", description="Enable Compiler Target for SYCL")
```
tomdeakin, line 78: I expect this will be pushed back on by the main Spack developers as there is another notation for choosing a compiler.
Can we detect this with the `%icpx` for Intel/`%syclcc`for OpenSYCL (hipSYCL)?

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+            if("cuda_arch" in spec.variants):
+                cuda_arch_list = spec.variants['cuda_arch'].value
+                # the architecture value is only number so append sm_ to the name
+                cuda_arch = "sm_" + cuda_arch_list[0]
+                args.append("-DOFFLOAD=ON" + " NVIDIA:" + cuda_arch)
+            elif("amdgpu_target" in spec.variants):
+                rocm_arch = spec.variants['amdgpu_target'].value
+                # the architecture value is only number so append sm_ to the name
+                args.append("-DOFFLOAD=ON" +" AMD:"+ rocm_arch)
+            # else:
```
tomdeakin, line 166: We should also support Intel OpenMP offload. You can test this on the NUC in the Zoo.

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+    
+    homepage = "https://github.com/UoB-HPC/BabelStream"
+    # url = "https://github.com/UoB-HPC/BabelStream/archive/refs/tags/v4.0.tar.gz"
+    # git = "https://github.com/UoB-HPC/BabelStream.git"
+    git = "https://github.com/kaanolgu/BabelStream.git" # this is temporary to have sycl2020 working before updates
+    # version("4.0", sha256="a9cd39277fb15d977d468435eb9b894f79f468233f0131509aa540ffda4f5953")
+    version('main', branch='main')
+    maintainers = ["robj0nes ", "tomdeakin", "tom91136"]
+    
+    #Languages
```
tomdeakin, line 27:
```suggestion
    #Languages
    # Also supported variants are cuda and rocm (for HIP)
```

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+
+        if "+tbb" in spec:
+            args.append("-DTBB_ROOT=" + self.spec["tbb"].prefix + "/lib/cmake/tbb")
+            # Hack to get around the fact that TBBConfig.cmake is linking 32bit libs.
+            args.append("-DCMAKE_SIZEOF_VOID_P=8")
+
+        #===================================
+        #             OpenCL (ocl) 
+        #===================================
+        if "+ocl" in spec:
```
tomdeakin, line 222: You might need to replicate some of the logic from CUDA to populate the `OpenCL_LIBRARY` variable with the right path based on if the OpenCL library is CUDA, HIP, Intel or POCL.

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+        #             kokkos 
+        #===================================
+
+        if "+kokkos" in spec:
+            args.append("-DCMAKE_CXX_COMPILER=" + self.compiler.cxx)
+            args.append("-DKOKKOS_IN_TREE=" + spec.variants['dir'].value)
+            args.append("-DKOKKOS_IN_PACKAGE=" + self.spec["kokkos"].prefix)
+            if self.spec.satisfies("+cuda"):
+                args.append("-DKokkos_ENABLE_CUDA=ON")
+            if self.spec.satisfies("+openmp"):
```
tomdeakin, line 252: Should this be `+omp` instead of `+openmp` to match the variant for normal OpenMP?

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+                args.append("-DKokkos_ENABLE_OPENMP=ON")
+
+        # not in ["kokkos", "raja", "acc", "hip"] then compiler forced true
```
tomdeakin, line 255: Maybe add a comment on why you need to do this?

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+                args.append("-DENABLE_OPENMP=ON")
+            if "+cuda" in spec:
+                args.append("-DENABLE_OPENMP=ON")
+    
+    
+        #===================================
+        #             kokkos 
+        #===================================
+
+        if "+kokkos" in spec:
```
tomdeakin, line 246: Can you add a comment saying "remember to set the arch via the usual kokkos method e.g., ..."?

apps/babelstream/README.md
```diff
@@ -0,0 +1,175 @@
+# Spack Package for BabelStream User Manual
+
+
+## Performance Results ( ISAMBARD Login-02 - Xeon E5-2695 v4)
```
tomdeakin, line 4: Can you split the results out to a separate file please?

tomdeakin (changes requested):
Lots of great stuff here @kaanolgu! This is a really good start.

I've been testing some stuff on the Zoo. I added some local packages in my `$HOME/.spack/packags.yaml` to prevent it installing LLVM, CUDA, etc.
CUDA works just great: `spack install babelstream +cuda cuda_arch=75`

OpenCL is set up to always uses POCL, which isn't what we want - we also want the OpenCL that comes with CUDA, HIP and Intel, and then install POCL as an option if needed/requested. Can you look again at how OpenCL works please?

Some other minor changes and suggestions throughout too.

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+
+class Babelstream(CMakePackage,CudaPackage,ROCmPackage):
+    """Measure memory transfer rates to/from global device memory on GPUs.
+    This benchmark is similar in spirit, and based on, the STREAM benchmark for CPUs."""
+
+    
+    homepage = "https://github.com/UoB-HPC/BabelStream"
+    # url = "https://github.com/UoB-HPC/BabelStream/archive/refs/tags/v4.0.tar.gz"
+    # git = "https://github.com/UoB-HPC/BabelStream.git"
+    git = "https://github.com/kaanolgu/BabelStream.git" # this is temporary to have sycl2020 working before updates
```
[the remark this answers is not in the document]
kaanolgu: This comment was from earlier so the requests you approved should be sufficient removing it

apps/babelstream/package.py
```diff
@@ -0,0 +1,268 @@
...
+    # version("4.0", sha256="a9cd39277fb15d977d468435eb9b894f79f468233f0131509aa540ffda4f5953")
+    version('main', branch='main')
```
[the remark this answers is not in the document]
kaanolgu: I am going to update this but keeping the quotes below from the original document for future reference

> Numeric versions are almost always tarball downloads, whereas non-numeric versions not named develop frequently download untrusted branches or tags from a version control system. As long as a package has at least one numeric version, and no non-numeric version named develop, Spack will prefer it over any non-numeric versions.

> If a package contains both a url and git class-level attribute, Spack decides which to use based on the arguments to the version() directive. Versions containing a specific branch, tag, or revision are assumed to be for VCS download methods, while versions containing a checksum are assumed to be for URL download methods.

apps/babelstream/package.py
kaanolgu: So something like this ? 
```
       if "+ocl" in spec:
            if("+cuda" in spec):
                cuda_dir = self.spec["cuda"].prefix
                cuda_comp = cuda_dir + "/bin/nvcc"
                args.append("-DCMAKE_CXX_COMPILER=" + cuda_comp)
                
                ````

File: apps/babelstream/package.py (hunk 8 above)
Comment (kaanolgu) [reply apps/babelstream/package.py:78]: I am afraid it is not possible to detect custom values with `%` symbol but is it OK to rename `compiler` to something like `compiler-option` ?

File: apps/babelstream/package.py (hunk 14 above)
Comment (kaanolgu) [reply apps/babelstream/package.py:246]: Kokkos for nvidia actually needs more flags to be set because it complains about arch flags need to be set for instance `Kokkos_ARCH_VOLTA72`  so I will code their conditions and flags here to

File: apps/babelstream/package.py (hunk 9 above)
Comment (kaanolgu) [reply apps/babelstream/package.py:166]: So the commented line actually at #L166 solves this I will uncomment it

**Comment by tomdeakin** [comment]
The BabelStream package is now in upstream Spack, so this PR will be repurposed to just merge in documentation for the benchmarking suite.

File: benchmarks/apps/babelstream/babelstream.py [hunk 18]
```diff
@@ -0,0 +1,220 @@
...
+    @performance_function('MBytes/sec', perf_key='Dot')
+    def extract_dot_perf(self):
+        return sn.extractsingle(r'Dot \s+(\S+)\s+.', self.stdout, 1, float)
+
+#--------------------------------------------------
+#-- ACC
+#--------------------------------------------------
+@rfm.simple_test
+class ACCBenchmark_CPU(BabelstreamBenchmarkBase):
+    valid_systems = ['-gpu']
```
giordano, line 75: Uhm, I just realised this will not work very well on Myriad where we have a single partition for CPU only and GPU jobs.  I'll need to think how to deal with this, I see two options:

* have two partitions for Myriad, but then in all CPU benchmarks we'll likely want to have `valid_systems = ['-gpu']`
* have a `cpu` feature for partitions

I like the latter option less, because in a sense pretty much all partitions will have CPUs, but the former option requires setting `valid_systems` explicitly everywhere.

kaanolgu: What if we remove the `gpu` from the features of Myriad 
https://github.com/ukri-excalibur/excalibur-tests/blob/4380722d0233438a615e0d64837339970546e712/benchmarks/reframe_config.py#L129
Or, 
Can we create another virtual `MyriadGPU` system in reframe config ?

giordano: > What if we remove the gpu from the features of Myriad

But it does have GPUs, they can be optionally requested.

>  Can we create another virtual MyriadGPU system in reframe config ?

That was my first option :slightly_smiling_face:

kaanolgu: Apologies :laughing: , yes I agree with the first option!

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,220 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+import os
+import os.path as path
+import sys
+import reframe as rfm
+import reframe.utility.sanity as sn
+from reframe.core.backends import getlauncher
+# Add top-level directory to `sys.path` so we can easily import extra modules
+# from any directory.
+sys.path.append(path.join(path.dirname(__file__), '..', '..'))
```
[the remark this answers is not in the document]
kaanolgu: It does not proceed to tests when I delete that line 
```
WARNING: skipping test file '/lustre/home/br-kolgu/excalibur-tests/benchmarks/apps/babelstream/babelstream.py': module not found error: benchmarks/apps/babelstream/babelstream.py:15: No module named 'modules'
from modules.utils import SpackTest
 (rerun with '-v' for more information)
```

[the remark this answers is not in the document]
giordano: None of the [other benchmarks](https://github.com/ukri-excalibur/excalibur-tests/tree/14820685caa7e3ad9131d1d8ec576d48a76ca02b/benchmarks/apps) use that anymore after #102.

tkoskela (changes requested):
Thanks for all the great work Kaan, we are nearly there!

We need a README in `benchmarks/apps/babelstream/` that documents the required ReFrame options. You can look at the other apps for examples, for example [cp2k](https://github.com/ukri-excalibur/excalibur-tests/tree/bristol_babelstream/benchmarks/apps/cp2k#readme)

tkoskela (approved):
The README looks good, thanks for adding it! I could run the tests on Isambard, so I'm happy to merge this in. I've had some issues building on other systems. These include

- On my laptop, building `babelstream%gcc+omp` errors due to unknown flags being passed to gcc. We discussed this and it needs to be fixed on the spack package side.
- On csd3 and cosma, the Intel compilers are missing paths to libraries I encountered at least `libsvm` and `libomptaget`. This needs to be fixed in our spack environments, I'll open a new issue.

tomdeakin (approved):

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,267 @@
...
+# The only requirement is to inherit this class and set the `spack_spec`
+# attribute.
+from benchmarks.modules.utils import SpackTest
+
+
+@rfm.simple_test
+class BabelstreamBenchmarkBase(SpackTest):
+    descr = 'Build BabelStream with Spack Build System'
+    valid_systems = ['*']
+    valid_prog_environs = ['default']
```
giordano, line 19:
```suggestion
    valid_prog_environs = ['default']
```

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,267 @@
...
+    valid_systems = ['*']
+    valid_prog_environs = ['default'] 
+
+    @run_before('run')
+    def replace_launcher(self):
+        # For this benchmark we don't use MPI at all, so we always force the
+        # local launcher:
+        # <https://reframe-hpc.readthedocs.io/en/v4.1.3/tutorial_advanced.html#replacing-the-parallel-launcher>.
+        self.job.launcher = getlauncher('local')()
+
```
giordano, line 27:
```suggestion

```

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,267 @@
...
+                'Dot': (0, None, 732000, 'MBytes/sec')
+            },
+            'isambard-macs:volta': {
+                'Copy':  (0, None, 897000, 'MBytes/sec'),
+                'Mul':  (0, None, 897000, 'MBytes/sec'),
+                'Add':  (0, None, 897000, 'MBytes/sec'),
+                'Triad':  (0, None, 897000, 'MBytes/sec'),
+                'Dot':  (0, None, 897000, 'MBytes/sec')
+                
+            },
```
giordano, line 102:
```suggestion
            },
```

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,267 @@
...
+            'isambard-phase3:ampere': {
+                'Copy':  (0, None, 1555000, 'MBytes/sec'),
+                'Mul':  (0, None, 1555000, 'MBytes/sec'),
+                'Add':  (0, None, 1555000, 'MBytes/sec'),
+                'Triad':  (0, None, 1555000, 'MBytes/sec'),
+                'Dot':  (0, None, 1555000, 'MBytes/sec')
+            },
+#             Myriad has four types of GPU nodes: E, F, J and L.
+# L-type nodes each have four NVIDIA 40G A100s. (Compute Capability 80)
+# F-type and E-type nodes each have two NVIDIA Tesla V100s. The CPUs are
```
giordano, line 112:
```suggestion
# F-type and E-type nodes each have two NVIDIA Tesla V100s. The CPUs are
```

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,220 @@
...
+    executable = "sycl-stream"
+
+#--------------------------------------------------
+#-- HIP(ROCm)
+#--------------------------------------------------
+@rfm.simple_test
+class HIPBenchmark(BabelstreamBenchmarkBase):
+    valid_systems = ['+gpu']
+    tags = {"hip"}
+    executable = "hip-stream"
```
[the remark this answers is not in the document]
giordano: :point_up:

.gitignore
```diff
@@ -21,3 +21,7 @@ outdated/
 
 # jupyter checkpoints:
 *-checkpoint.ipynb
+
+#ignore virtual environment
+myvenv/
+perflogs/
```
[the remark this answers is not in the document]
giordano: :point_up:

giordano (commented):
A few whitespace issues

tkoskela:
Hi, we got the Intel builds working on a few different systems with #177. I noticed that the `OMPBenchmark_CPU` only runs on 1 core. Is it supposed to do that? What's the OMP doing?

kaanolgu:
That's a nice point for the CPU benchmarks for pointing out this @tkoskela , I might need to add instructions to the ReFrame readme too for this.

The number of CPU cores are not defined hence it uses only 1 core since we don't know which machine is going to be chosen by user and the purpose, user might need to only test if it works or not. So I didn't specify the number of cpus in the code but user can always change the values by using the similar approach done here

https://github.com/ukri-excalibur/excalibur-tests/blob/2a01b2c694f6bdd29f097e18c5060fbb44988e6c/benchmarks/apps/wrf/README.md?plain=1#L42C1-L50C20

## Let's test it ! with default settings 
`reframe -c benchmarks/apps/babelstream -r --tag omp --system=isambard-macs:cascadelake -S build_locally=false -S spack_spec='babelstream%oneapi +omp'`

```
[----------] start processing checks
[ RUN      ] OMPBenchmark_CPU /ebd97f73 @isambard-macs:cascadelake+default
[       OK ] (1/1) OMPBenchmark_CPU /ebd97f73 @isambard-macs:cascadelake+default
P: Copy: 12714.124 MBytes/sec (r:0, l:None, u:None)
P: Mul: 12904.714 MBytes/sec (r:0, l:None, u:None)
P: Add: 13311.646 MBytes/sec (r:0, l:None, u:None)
P: Triad: 13328.291 MBytes/sec (r:0, l:None, u:None)
P: Dot: 13896.437 MBytes/sec (r:0, l:None, u:None)
[----------] all spawned checks have finished
```
job file has the following: 
```
#!/bin/bash -l
#PBS -N rfm_job
#PBS -o rfm_job.out
#PBS -e rfm_job.err
#PBS -l walltime=0:15:0
#PBS -l select=1:mpiprocs=1:ncpus=1
#PBS -q clxq
cd /lustre/home/br-kolgu/excalibur-tests/stage/isambard-macs/cascadelake/default/OMPBenchmark_CPU
eval `spack -e /lustre/home/br-kolgu/excalibur-tests/stage/isambard-macs/cascadelake/default/OMPBenchmark_CPU/spack_env/cascadelake load --sh babelstream%oneapi +omp`
omp-stream

```

## Now let's test setvar cpu to 4 :
`reframe -c benchmarks/apps/babelstream -r --tag omp --system=isambard-macs:cascadelake --setvar=num_cpus_per_task=4 -S build_locally=false -S spack_spec='babelstream%oneapi +omp'`

```
[----------] start processing checks
[ RUN      ] OMPBenchmark_CPU /ebd97f73 @isambard-macs:cascadelake+default
[       OK ] (1/1) OMPBenchmark_CPU /ebd97f73 @isambard-macs:cascadelake+default
P: Copy: 47774.028 MBytes/sec (r:0, l:None, u:None)
P: Mul: 47977.02 MBytes/sec (r:0, l:None, u:None)
P: Add: 49577.994 MBytes/sec (r:0, l:None, u:None)
P: Triad: 49530.413 MBytes/sec (r:0, l:None, u:None)
P: Dot: 50863.703 MBytes/sec (r:0, l:None, u:None)
[----------] all spawned checks have finished
```

Job script :
```
#!/bin/bash -l
#PBS -N rfm_job
#PBS -o rfm_job.out
#PBS -e rfm_job.err
#PBS -l walltime=0:15:0
#PBS -l select=1:mpiprocs=1:ncpus=4
#PBS -q clxq
cd /lustre/home/br-kolgu/excalibur-tests/stage/isambard-macs/cascadelake/default/OMPBenchmark_CPU
eval `spack -e /lustre/home/br-kolgu/excalibur-tests/stage/isambard-macs/cascadelake/default/OMPBenchmark_CPU/spack_env/cascadelake load --sh babelstream%oneapi +omp`
omp-stream

```
So the results are ~4 times so it works!

benchmarks/apps/babelstream/babelstream.py
giordano: For the record, the first option had been implemented in #164.

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,261 @@
...
+@rfm.simple_test
+class THRUSTBenchmark_NVIDIA(BabelstreamBenchmarkBase):
+    valid_systems = ['+gpu +cuda']
+    tags = {"thrust"}
+    executable = "thrust-stream"
+    num_gpus_per_node = 1
+
+@rfm.simple_test
+class THRUSTBenchmark_AMD(BabelstreamBenchmarkBase):
+    valid_systems = ['+gpu']
```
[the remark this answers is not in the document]
kaanolgu: > 

Thanks for mentioning I will add it to `_AMD` ones but is it better to use `+rocm` or `-cuda` ?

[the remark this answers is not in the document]
giordano: My understanding is that this requires AMD GPUs, so an explicit `+rocm` would be better, to avoid confusion when we'll ever have Intel (or other vendors) GPUs

benchmarks/apps/babelstream/babelstream.py
```diff
@@ -0,0 +1,261 @@
...
+    tags = {"tbb"}
+    executable = "tbb-stream"
+    def __init__(self):
+        self.reference = {
+            'isambard-macs:cascadelake': { # https://en.wikichip.org/wiki/intel/xeon_gold/6230https://en.wikichip.org/wiki/intel/xeon_gold/6230 (SINGLE)
+                'Copy':  (0, None, 231860, 'MBytes/sec'),
+                'Mul': (0, None, 231860, 'MBytes/sec'),
+                'Add':   (0, None, 231860, 'MBytes/sec'),
+                'Triad': (0, None, 231860, 'MBytes/sec'),
+                'Dot': (0, None, 231860, 'MBytes/sec')
```
giordano, line 169: Are these reference values correct?  The first element of the tuple should be the value, the second and the third ones the lower and upper bounds, as a fraction of the value if not `None`, but I'm not sure fractions of `0` are useful, but I suspect you wanted to do something like
```py
(231860, None, None, 'MBytes/sec'),
```
instead, although I don't know if we want to have reference values at all here, given previous discussions.

kaanolgu:
> > Regarding OpenMP, don't you need to set `OMP_NUM_THREADS` to actually use the threads? On CSD3, with the current PR I get:
> > ```
> > $ reframe -c benchmarks/apps/babelstream -r --system csd3-skylake:compute-node --job-option='--account=DIRAC-DO006-CPU' -S spack_spec='babelstream%intel +omp' --tag omp
> > [...]
> > P: Copy: 12538.703 MBytes/sec (r:0, l:None, u:None)
> > P: Mul: 12632.656 MBytes/sec (r:0, l:None, u:None)
> > P: Add: 12959.008 MBytes/sec (r:0, l:None, u:None)
> > P: Triad: 12964.855 MBytes/sec (r:0, l:None, u:None)
> > P: Dot: 10230.426 MBytes/sec (r:0, l:None, u:None)
> > [...]
> > $ reframe -c benchmarks/apps/babelstream -r --system csd3-skylake:compute-node --job-option='--account=DIRAC-DO006-CPU' -S spack_spec='babelstream%intel +omp' --tag omp --setvar=num_cpus_per_task=4
> > [...]
> > P: Copy: 12323.726 MBytes/sec (r:0, l:None, u:None)
> > P: Mul: 12508.92 MBytes/sec (r:0, l:None, u:None)
> > P: Add: 12790.734 MBytes/sec (r:0, l:None, u:None)
> > P: Triad: 12777.699 MBytes/sec (r:0, l:None, u:None)
> > P: Dot: 10043.903 MBytes/sec (r:0, l:None, u:None)
> > [...]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but with a patch like
> > ```diff
> > diff --git a/benchmarks/apps/babelstream/babelstream.py b/benchmarks/apps/babelstream/babelstream.py
> > index 139b64e..0e2c87a 100644
> > --- a/benchmarks/apps/babelstream/babelstream.py
> > +++ b/benchmarks/apps/babelstream/babelstream.py
> > @@ -54,6 +54,11 @@ class BabelstreamBenchmarkBase(SpackTest):
> >      def extract_dot_perf(self):
> >          return sn.extractsingle(r'Dot \s+(\S+)\s+.', self.stdout, 1, float)
> >  
> > +    @run_after('setup')
> > +    def setup_num_tasks(self):
> > +        self.env_vars['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
> > +
> > +
> >  #--------------------------------------------------
> >  #-- ACC
> >  #--------------------------------------------------
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead I get
> > ```
> > $ reframe -c benchmarks/apps/babelstream -r --system csd3-skylake:compute-node --job-option='--account=DIRAC-DO006-CPU' -S spack_spec='babelstream%intel +omp' --tag omp --setvar=num_cpus_per_task=4
> > [...]
> > P: Copy: 46661.417 MBytes/sec (r:0, l:None, u:None)
> > P: Mul: 46605.166 MBytes/sec (r:0, l:None, u:None)
> > P: Add: 48610.452 MBytes/sec (r:0, l:None, u:None)
> > P: Triad: 48631.973 MBytes/sec (r:0, l:None, u:None)
> > P: Dot: 38978.97 MBytes/sec (r:0, l:None, u:None)
> > [...]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which is closer to the advertised 4x speedup.
> 
> Interesting, It might be the case that I am on the Spack package PR [spack/spack#38559](https://github.com/spack/spack/pull/38559) and using no offload with intel compilers or there might be a system related interpretation difference ? In order to be more consistent I will add the changes you made to test

I did update my main branch locally where I was a bit behind but still can't seem to reproduce the same issue you pointed out on `isambard-macs:cascadelake` so with or without the line you suggested it makes the 4x improvement for me

kaanolgu:
> I wonder if that depends on the scheduler and relying on that instead of `OMP_NUM_THREADS` is a good idea.

I did add it to the recent PR thanks for pointing out and did change the values for the TBB they were just there for the perflog to see the values since as you said we agreed to not to put the reference values there for babelstream

PR #99, opened 2023-02-27
Delete old perflogs

giordano:
They aren't useful for anything for our purposes at the moment and only clutter the repository.  The diff in #91 is totally unreadable because of all these files.


tkoskela (approved):
Yeah, makes sense to me. We weren't planning to store perflogs anyways, and we don't even know how these were made.

ilectra (approved):

PR #111, opened 2023-03-08
Upgrade to ReFrame v4

### LINKED ISSUE #89: Consider moving to ReFrame v4

giordano:
Haven't tested thoroughly, but the simple sombrero example works for me for the time being.

Eventually will fix #89.

### DIFF
--- a/benchmarks/reframe_config.py
+++ b/benchmarks/reframe_config.py
@@ -34,7 +34,7 @@ site_configuration = {
                     'descr': 'ARCHER2 compute nodes',
                     'scheduler': 'slurm',
                     'launcher': 'srun',
-                    'variables': [
+                    'env_vars': [
                         # Propagate PATH to compute nodes, including `spack` bindir
                         ['PATH', spack_root_to_path()],
                         # Work around for Spack erroring out on non-existing home directory:
@@ -213,6 +213,9 @@ site_configuration = {
                     'launcher': 'mpiexec',
                     'access': ['--partition=cosma8'],
                     'environs': ['default', 'intel20-mpi-durham', 'intel20_u2-mpi-durham', 'intel19-mpi-durham', 'intel19_u3-mpi-durham'],
+                    'sched_options': {
+                        'use_nodes_option': True,
+                    },
                     'max_jobs': 64,
                     'processor': {
                         'num_cpus': 128,
@@ -271,6 +274,9 @@ site_configuration = {
                     'launcher': 'mpirun',
                     'access': ['--partition=cpu', '--qos=standard'],
                     'environs': ['default'],
+                    'sched_options': {
+                        'use_nodes_option': True,
+                    },
                     'max_jobs': 16,
                     'processor': {
                         'num_cpus': 64,
@@ -432,18 +438,11 @@ site_configuration = {
                         'upper=%(check_perf_upper_thres)s|'
                         'units=%(check_perf_unit)s|'
                         'spack_spec=%(check_spack_spec)s|'
-                        'variables=%(check_variables)s'
+                        'env_vars=%(check_env_vars)s'
                     ),
                     'append': True
                 }
             ]
         }
-    ],
-    'schedulers': [
-        {
-            'name': 'slurm',
-            'target_systems': ['tursa', 'cosma8'],
-            'use_nodes_option': True,
-        },
     ],
 }
--- a/benchmarks/reframe_config.py
+++ b/benchmarks/reframe_config.py
@@ -429,16 +429,19 @@ site_configuration = {
                         'reframe %(version)s|'
                         '%(check_info)s|'
                         'jobid=%(check_jobid)s|'
-                        '%(check_perf_var)s=%(check_perf_value)s|'
                         'num_tasks=%(check_num_tasks)s|'
                         'num_cpus_per_task=%(check_num_cpus_per_task)s|'
                         'num_tasks_per_node=%(check_num_tasks_per_node)s|'
-                        'ref=%(check_perf_ref)s|'
-                        'lower=%(check_perf_lower_thres)s|'
-                        'upper=%(check_perf_upper_thres)s|'
-                        'units=%(check_perf_unit)s|'
+                        '%(check_perfvalues)s|'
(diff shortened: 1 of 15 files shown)


giordano:
`sched_options` may not be available in v4.0.0 despite what the [documentation](https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#config.systems.partitions.sched_options) says.

reframe_config.py
```diff
@@ -432,18 +438,11 @@ def spack_root_to_path():
                         'upper=%(check_perf_upper_thres)s|'
                         'units=%(check_perf_unit)s|'
                         'spack_spec=%(check_spack_spec)s|'
-                        'variables=%(check_variables)s'
+                        'env_vars=%(check_env_vars)s'
```
ilectra, line 441: The `check_perf_*` fields above probably have to be modified as well, according to https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#config.logging.handlers_perflog.format_perfvars . Or turn on compatibility with `logging.perflog_compat`

giordano: Thanks, done.  Example of new perflog:
```
job_completion_time|version|info|jobid|num_tasks|num_cpus_per_task|num_tasks_per_node|flops_var=flops_value|flops_ref|flops_lower|flops_upper|flops_unit|spack_spec|env_vars
2023-03-08T17:38:06|reframe 4.1.0|SombreroBenchmark /cc391b86 @generic:default+default|jobid=74922|num_tasks=1|num_cpus_per_task=1|num_tasks_per_node=1|flops=1.15|ref=1|lower=None|upper=None|units=Gflops/seconds|spack_spec=sombrero@2021-08-16|env_vars={"OMP_NUM_THREADS": "1"}
2023-03-08T17:38:58|reframe 4.1.0|SombreroBenchmark /cc391b86 @generic:default+default|jobid=75084|num_tasks=1|num_cpus_per_task=1|num_tasks_per_node=1|flops=1.16|ref=1|lower=None|upper=None|units=Gflops/seconds|spack_spec=sombrero@2021-08-16|env_vars={"OMP_NUM_THREADS": "1"}
2023-03-08T17:40:41|reframe 4.1.0|SombreroBenchmark /cc391b86 @generic:default+default|jobid=75511|num_tasks=1|num_cpus_per_task=1|num_tasks_per_node=1|flops=1.12|ref=1|lower=None|upper=None|units=Gflops/seconds|spack_spec=sombrero@2021-08-16|env_vars={"OMP_NUM_THREADS": "1"}
```

giordano:
@TomMelt since you independently worked on this, is there anything that we're missing here from what you've done in #116?

TomMelt:
> @TomMelt since you independently worked on this, is there anything that we're missing here from what you've done in #116?

On first inspection no, it looks like you covered everything... however I have some style differences in the perf logs

image

In this version we still have the old fashioned `env_vars=%(check_env_vars)s` for example. But the new perf logs have a header row which describes each column. Therefore I would vote to remove this redundant information. Is there a reason you are keeping it?

benchmarks/reframe_config.py
```diff
@@ -423,27 +429,23 @@ def spack_root_to_path():
...
+                    'format_perfvars': (
+                        '%(check_perf_var)s=%(check_perf_value)s|'
                         'ref=%(check_perf_ref)s|'
-                        'lower=%(check_perf_lower_thres)s|'
-                        'upper=%(check_perf_upper_thres)s|'
+                        'lower=%(check_perf_lower)s|'
+                        'upper=%(check_perf_upper)s|'
                         'units=%(check_perf_unit)s|'
-                        'spack_spec=%(check_spack_spec)s|'
-                        'variables=%(check_variables)s'
```
TomMelt, line 435: Specifically in these lines. Do we really need the `XXXXX=%(XXXXX)`. The new perf log header will print this information anyway

giordano: Good point.  I'll defer the decision to @ilectra and @pineapple-cat, as they are more interested in the exact format of the perflogs, but I think you're right that we should be able to get rid of them now.

giordano: And I agree also with changing `%(check_perf_var)s=%(check_perf_value)s` to `%(check_perf_value)s`.

pineapple-cat: I think it might be worth having separate columns for variables and values like so `perf_var=%(check_perf_var)s|perf_val=%(check_perf_value)s` because we can have multiple figures of merit, though I'm not sure what the default format is for displaying the variables and values when there are more than one. Aside from that, feel free to push this since I'll be making more changes to the format anyway, starting with getting all fields to conform to a `name=value` pattern.

pineapple-cat (approved):

TomMelt:
Hi @pineapple-cat , perhaps I didn't understand correctly. But, you don't need a separate column for variable name and values. The new format looks something like this, where I have 3 perf variables:
* diagonalization
* test
* energy

```
job_completion_time | version       | info                                                                        | jobid  | num_tasks | num_cpus_per_task | num_tasks_per_node | diagonlization_value | diagonlization_unit | diagonlization_ref | diagonlization_lower | diagonlization_upper | test_value | test_unit | test_ref | test_lower | test_upper | Energy_value | Energy_unit | Energy_ref | Energy_lower | Energy_upper | spack_spec                                           | variables
2023-03-30T14:33:07 | reframe 4.1.0 | PdsyevSingle %num_threads=2 /e04d8cdd @dial3:compute-node+intel19-mpi-dial3 | 169766 | 64        | null              | 64                 | 1.58602              | seconds             | 200                | None                 | None                 | 1.58602    | seconds   | 250      | None       | None       | 0            | Wh          | 0          | None         | None         | trove-pdsyev%intel@19.0.5.281 ^intel-parallel-studio | {"OMP_NUM_THREADS": "2", "OMP_PLACES": "cores"}
```

The name of the perf variables in included in the header e.g., `diagonlization_value`,  `test_value` etc.

pineapple-cat:
Aha, I see now. In that case this looks good to me; I just wanted to make sure there was a way to identify how many figures of merit there are in a perflog, along with their names, values, units etc.

ilectra (approved):

PR #114, opened 2023-03-13
Add docs on requesting GPU nodes on CSD3

tkoskela:
Here I've assumed we define different partitions for cpu and gpu jobs on the same system, and called the gpu partition in the example `gpu-node`. Open to changes on this.


giordano:
> Here I've assumed we define different partitions for cpu and gpu jobs on the same system, and called the gpu partition in the example `gpu-node`.

Should add said partition in the same PR? :grin:

tkoskela:
Can do, since I've been playing with it :grin:

tkoskela:
Yeah we can probably close this PR if all the user has to do is pick the right partition and ReFrame will handle the rest.

PR #116, opened 2023-03-22
Update leicester codes

TomMelt:
Asif and I have modified Lokesh's test classes to work with spack

We have also updated them to work with reframe v>=4.0

We have moved them out from the dirac folder to the same level as the other apps.

We have tested the new classes on our system here at Leicester and they produce similar results to the ones Lokesh ran before.

There is still some work to do, but this contains the most important changes.

Further work includes:
* moving from University of Leciester private github repos (see corresponding spack recipes) to the repos managed by respective PIs (I will submit a PR for our spack recipes in the next week).
* Expanding tests to run with more compilers. Currently they have only been tested for intel. We think some of the codes will require source code changes to run with gnu compilers.


TomMelt:
> First of all, it's great to see this coming together! However, a quick comment: I'd prefer to keep logically distinct changes separated. For example, upgrade to reframe 4 is already part of #111 (which is waiting on #102 to avoid conflicts, but the latter should be very close to be merged).
> 
> Heads-up: #102 will reorganise the tree structure of the benchmarks a bit, but also simplify some stuff: installation of dependencies will be more straightforward, and we don't need to use the horrible `sys.path` hack anymore to access the `SpackTest` module.

That's great to hear. Perhaps I will wait for #111 and #102 to be merged and then I can tidy our changes up a little bit.

TomMelt:
Hi @giordano , do you have any update on the progress for #111 ?

Should I make a branch off of `mg/reframe-v4` and make the changes there? Or should I branch off `main`?

In your previous comment you said:
> First of all, it's great to see this coming together! However, a quick comment: I'd prefer to keep logically distinct changes > separated. For example, upgrade to reframe 4 is already part of #111

I did create a separate branch for upgrading to reframe 4 (`TomMelt:update-for-reframev4`) which I then created the branch `TomMelt:update-leicester-codes` from. However, maybe because I submitted the merge from my fork, the information is lost as the other branch doesn't exist here.

Unfortunately, we can't really separate our new tests from the update because our new tests rely on reframe v4 features. The old ones don't work at all. Let me know what you think is best :+1:

TomMelt:
I am replacing this PR with #131 due to big changes in the layout of the repo.

Issue #105, opened 2023-03-08
Read perflog data into pandas dataframe


ilectra:
Inter-connected tasks needed:
- [x] Get the [handlers_perflog.format](https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#config.logging.handlers_perflog.format) in https://github.com/ukri-excalibur/excalibur-tests/blob/87dde005a6b581448a9003c8dcbac47107019957/reframe_config.py#L393 to
  - [x] follow the pattern of `name=value` for every field
  - [x] add all the needed variables: `check_variables` dictionary of environment variables, tags, either `check_info` or `display_name` - see https://github.com/ukri-excalibur/excalibur-tests/issues/70#issuecomment-1453558145 and be aware of ReFrame version.
  - [x] add info encoded in the perflog file path (system, partition) as fields inside the perflog as well
- [x] write perflog parser to be as generic as possible, assuming `name=value` format of fields. Use [existing code](https://github.com/ukri-excalibur/excalibur-tests/blob/d745085064a0ad1bd351fdf1ff8229ab13b4cbe2/modules/perf_logs.py) where possible.
  - [x] input perflog files can be coming from various tests of one app, various systems, or even more than one app, depending on the usecase. This will modify the way the input path will be searched for files.
  - [x] Fishing out the parameters values from the `info` or `display_name` field will need its own parsing.

PR #121, opened 2023-03-27
Add perflog parsing for post-processing

pineapple-cat:
Addresses #105.

- [x] Perflog parsing
- [x] Display name parsing
- [x] Unit tests for perflog formatting and parsing


benchmarks/examples/sombrero/sombrero.py
```diff
@@ -100,6 +96,19 @@ class SombreroBenchmark(SpackTest):
         },
     }
 
+    def __init__(self):
```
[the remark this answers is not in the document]
pineapple-cat: I tried it without but it just skips the file because of the following: `reframe.core.exceptions.ReframeSyntaxError: accessing a test parameter from the class body is disallowed`. I took the `__init__` idea from the IMB benchmarks.

benchmarks/examples/sombrero/sombrero.py
```diff
@@ -100,6 +96,19 @@ class SombreroBenchmark(SpackTest):
...
+        # accessing a test parameter from the class body is not allowed.
+        self.num_tasks = self.tasks
+        self.num_cpus_per_task = self.cpus_per_task
+        # Tags are useful for categorizing tests and quickly selecting those of interest.
+        self.tags.add("example")
+        self.tags.add("test" + str(self.tasks - 2 + self.cpus_per_task * 2))
+        # These extra resources are needed for when using the SGE scheduler.
+        self.extra_resources = {
+            'mpi': {'num_slots': self.num_tasks * self.num_cpus_per_task}
+        }
```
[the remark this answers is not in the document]
pineapple-cat: Aha, I'll go ahead and delete this line then.

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # look for existing log files
+    sombrero_logs = []
+    sombrero_log_re = re.compile(r"^SombreroBenchmark\_\w{8}\.log$")
+    for subdirs, dirs, files in os.walk(perflog_path):
+        sombrero_logs = list(filter(sombrero_log_re.match, files))
+
+    try:
+        # to save time don't re-generate test log files if they already exist
+        if not sombrero_logs:
```
ilectra, line 77: Remove and regenerate logs in the beginning of the test every time.
The optional deletion at the end of the test is fine as is.

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,132 @@
...
+            # use sombrero example to generate new test perflog file
+            run_benchmark(sombrero_bench_path)
+            benchmark_run = True
+            # look for log files again
+            for subdirs, dirs, files in os.walk(perflog_path):
+                sombrero_logs = list(filter(sombrero_log_re.match, files))
+        # check that the log files exist
+        if sombrero_logs:
+            # arbitrarily pick one of the log file names
```
ilectra, line 86: Check that you get the expected number of perflog files, then choose one arbitrarily.

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,132 @@
...
+            assert False
+
+        # get dataframe from complete perflog
+        df = post.read_perflog(sombrero_log_path)
+
+    finally:
+        # clean unnecessary files and folders
+        benchmark_cleanup(benchmark_run, remove_test_logs)
+
+    # check example perflog file is read appropriately
```
ilectra, line 126: Move those asserts that check on the df, up to line121

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,132 @@
...
+    assert df.columns.tolist() == EXPECTED_FIELDS
```
ilectra, line 127: Also check the content of the columns. Some of them will be expected to be empty.

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,132 @@
...
+    post_processing_dir = Path(__file__).parent
+    # change current working dir to avoid
+    # potentially deleting non-test perflogs later
+    os.chdir(post_processing_dir)
+    # resolve relative path from post-processing directory to access sombrero benchmark
+    sombrero_bench_path = str((post_processing_dir / "../benchmarks/examples/sombrero/sombrero.py").resolve())
+
+    # get test log file paths
+    perflog_path = os.path.join(os.getcwd(),"perflogs/default/default/")
+    sombrero_incomplete_log = "SombreroBenchmarkIncomplete.log"
```
ilectra, line 66: Move everything related to `sombrero_incomplete_log` down to where it's used (line 92)

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,132 @@
...
+    sombrero_incomplete_log_path = os.path.join(perflog_path, sombrero_incomplete_log)
+
+    # look for existing log files
```
ilectra, line 69: Would we gain anything here if we used the search in https://github.com/ukri-excalibur/excalibur-tests/blob/main/benchmarks/modules/utils.py#L99, called by https://github.com/ukri-excalibur/excalibur-tests/blob/main/benchmarks/modules/perf_logs.py#L72 ?

ilectra (commented):
A few more comments from me, I'm done with my review. 
Could you also add the unit tests to the GitHub actions? It "should" be straightforward, you've done the hard part already. It's probably just adding a line at the end of https://github.com/ukri-excalibur/excalibur-tests/blob/main/.github/workflows/ci.yml . If you get stuck, don't wait too long before you bug @giordano  - he needs to be kept busy :)

pineapple-cat:
Implemented suggestions:

- [ ] Restructure code for easier reading
- [ ] Regenerate perflogs each time the test is run
- [ ] Add extra assert statements to check number of perflogs + DataFrame column contents
- [ ] Simplify tags used in Sombrero example
- [ ] Add parameters to DataFrame
- [ ] Add tests to GitHub actions

post-processing/test_post_processing.py
pineapple-cat: Code reuse would be nice, but I can't seem to get this to work with the regex I'd like to match my log file names to.

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # check the log files exist
+    if sombrero_logs:
+        # check the sombrero log path is valid
+        assert os.path.exists(sombrero_log_path)
+        # check the expected number of log files has been generated
+        assert len(sombrero_logs) ==  4
+    # if the log files haven't been created, something went wrong
+    else:
+        assert False
```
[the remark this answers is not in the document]
pineapple-cat: It's difficult to put an error message here because the actual problem occurs in `run_sombrero`. I could throw/print some exceptions in the fixture if that would help, I just haven't done so because running `pytest -s` prints everything that happens when `reframe` is run to console, which is usually enough information for me to figure out what's wrong. I think on a clean installation a common issue could be that the `reframe_config.py` settings file wasn't found.

tkoskela (approved):
Looks good to me now

tkoskela (approved):
Looks good to me now

post-processing/test_post_processing.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # if the incomplete log file hasn't been created, something went wrong
+    if not os.path.exists(sombrero_incomplete_log_path):
```
ilectra, line 135: Why not `assert(os.path.exists(sombrero_incomplete_log_path))` instead of the 2-line `if` statement?
Same comment for anywhere else you use `assert False` (but not the check that the exception is thrown below, that one is fine).

ilectra (commented):
Almost there :)

PR #123, opened 2023-04-03
[Tursa] Automatically set `--account` flag for scheduler

giordano:
On Tursa we can automatically set the `--account` flag, unfortunately on the other DiRac systems I could test `reframe.utility.osext.osgroup()` doesn't return what we need.

### DIFF
--- a/SYSTEMS.md
+++ b/SYSTEMS.md
@@ -137,16 +137,3 @@ reframe -c examples/sombrero -r --performance-report --system tesseract:compute-
 ```
 
 where `<ACCOUNT>` is the project you want to charge.
-
-## Tursa
-
-### Queue options
-
-When submitting jobs to compute nodes, you need to specify the job queue, with the `--account` option to the scheduler.
-To do this, when you run a benchmark you can use the `-J`/`--job-option` flag to `reframe` to specify the account, for example:
-
-```
-reframe -c examples/sombrero -r --performance-report --system tursa:compute-node -J'--accout=<ACCOUNT>'
-```
-
-where `<ACCOUNT>` is the project you want to charge.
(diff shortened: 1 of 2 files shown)


SYSTEMS.md
```diff
@@ -137,16 +137,3 @@ reframe -c examples/sombrero -r --performance-report --system tesseract:compute-
...
-### Queue options
-
-When submitting jobs to compute nodes, you need to specify the job queue, with the `--account` option to the scheduler.
-To do this, when you run a benchmark you can use the `-J`/`--job-option` flag to `reframe` to specify the account, for example:
-
-```
-reframe -c examples/sombrero -r --performance-report --system tursa:compute-node -J'--accout=<ACCOUNT>'
-```
-
-where `<ACCOUNT>` is the project you want to charge.
```
tkoskela, line 152: How confident are you that this will always work? Would it be better to keep this section but rephrase it as "We will try to automatically detect your account, but in case this fails..."

tkoskela (commented):
The documentation of `osgroup()` is not entirely clear but it feels like it returns the first group the user belongs to? This seems a bit fragile to me.

What happens in ReFrame if `osgroup()` returns `None`? I guess the expression will evaluate to nothing and submitting the job will fail because account is not set. Does setting it on the command line still override the reframe config?

SYSTEMS.md
giordano: This is what they do in [CSCS Piz Daint reframe benchmarks](https://github.com/reframe-hpc/cscs-reframe-tests/blob/10ac6b5ac0a08c1909fee742fccd0b87ef4610cd/config/systems/daint.py#L49), where I took inspiration for this.  Sounds like they're highly confident this might work?

giordano:
Uhm, judging by https://epcced.github.io/dirac-docs/tursa-user-guide/scheduler/#checking-available-budget I guess a user may be on multiple budgets, so this isn't going to work.

PR #115, opened 2023-03-13
[OpenMM] Add new benchmark

giordano:
Part of #98.  Based on https://github.com/UCL-RITS/HPC-Acceptance-Tests/tree/681a51be96601c77f5c3106e8deba0c768cf46f4/Applications/OpenMM.  I tried it on Myriad with
```
reframe -c apps/openmm -r --performance-report --system myriad:compute-node --setvar spack_spec='openmm@7.7.0 ^openmpi'
```
It works in the sense that the package can be compiled with Spack successfully (after https://github.com/spack/spack/pull/35912 was merged) and input files are downloaded correctly, but I haven't been able yet to run the program through the end, which also means there's no performance metric, as I don't know yet what the figure of merit is at all.

Opening as draft, to be merged after #102.

PR #124, opened 2023-04-13
[reframe] Use new names of performance reference thresholds

giordano:
Names were uniformised in ReFrame v4.1.3.  These are the names that are now [documented](https://reframe-hpc.readthedocs.io/en/v4.1.3/config_reference.html#config.logging.handlers_perflog.format_perfvars).


giordano:
Ok, I'm going ahead and merge this because this change is ***necessary*** with ReFrame v4.1.3, I tested it with #115 on Tursa.

PR #128, opened 2023-04-14
Add COSMA7 partitions, including RockPort

giordano:


giordano:
I did some quick tests and it seems to be working well.

PR #131, opened 2023-04-19
Update leicester benchmarks

TomMelt:
This pull request replaces #116.

Due to some big changes to the structure of this repo it was easier to just start again.

Main differences are:
* I have now added spack recipes to the repo folder. This is because spack would not accept private github repos.
* ramses input data is now downloaded automatically as part of the test. (stored on [zenodo](https://zenodo.org/record/7842140/) thanks to suggestion from @giordano )
* tests import spacktest as a module now

@asifsamiarain and I still have a few checks to carryout but we should be ready next week.


TomMelt:
Asif and I have tested and it seems to be working fine

benchmarks/apps/ramses/ramses.py
```diff
@@ -0,0 +1,93 @@
...
+    def validate_download(self):
+        return sn.assert_true(path.exists('data.tgz'))
+
+
+class RamsesMPI(SpackTest):
+
+    valid_systems = ['*']
+    valid_prog_environs = ['*']
+
+    spack_spec = 'ramses@v1.0.0'
```
giordano, line 26: As far as I can tell, `ramses` and all the other packages below aren't in Spack, so how is this supposed to work?  If you have your own Spack recipes and are willing to keep them public in this repository while the software remain private for the time being, you could add them to our [custom spack repo](https://github.com/ukri-excalibur/excalibur-tests/tree/6f3cfbd4af8c031986e23af9f8da6360dfee36c5/benchmarks/spack/repo).

giordano (commented):
Looks overall good to me, I only left a couple of comments, but couldn't test myself.  Thanks!

benchmarks/apps/ramses/ramses.py
TomMelt: That's the point of this merge request ;) we have added them to the custom spack repo. You can see if you visit our fork / or if you look in the 'files changed' tab. Very happy for them to be public but they cannot be merged into spack as they are private github repos.

image

benchmarks/apps/sphng/sphng.py
```diff
@@ -0,0 +1,200 @@
...
+
+        self.thread_count = str(int(self.core_count_1_node/self.num_tasks_per_node))
+
+        self.descr = ('Running Sphng (Ifile) on ' + str(self.num_tasks_per_node) +' tasks and ' + \
+                       self.thread_count +  ' threads per node')
+
+        self.env_vars= {
+            'OMP_NUM_THREADS':self.thread_count,
+            'OMP_PLACES':'cores'
+        }
```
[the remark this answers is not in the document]
TomMelt: > Looks overall good to me, I only left a couple of comments, but couldn't test myself. Thanks!

I will add you with read access to the github repos and then you should be able to run. Wouldn't hurt to get an external validation of these tests. In principle I don't see any problem adding anyone who wants to run the test. But perhaps a discussion we should have in the DiRAC meeting.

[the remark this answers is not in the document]
TomMelt: > Here and in all other places below:
> 
> so that we don't completely [override the `env_vars` attribute](https://github.com/ukri-excalibur/excalibur-tests/pull/111/commits/94bacc1d67f9ad91f5cbf4caf4ce03b783a4ac1c).

Good spot. I will change these later today.

benchmarks/apps/ramses/ramses.py
giordano: Sorry, I completely missed those files at the end :facepalm:

benchmarks/spack/repo/packages/trove/package.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if self.compiler.name == "intel":
+            fflags = " -mavx2 -mfma -O3 -ip -Ofast"
+            if "openmpi" in spec:
+                lapack = (
+                    "-mkl=parallel -lmkl_scalapack_lp64 -lmkl_blacs_openmpi_lp64 -lmkl_intel_lp64"
+                )
+
+        elif self.compiler.name == "gcc":
+            fflags = " -ffree-line-length-none -march=native -O3 " "-fcray-pointer -g3"
```
giordano, line 47: Note that in general spack internally optimises compilation for the requested target, there's usually no need to use `-march=native`, which also prevents correct cross-compilation for a different ISA.

tkoskela:
If there's any system-specific tricks required on Dial2 or Dial3 outside of the ReFrame script, could you please add them to SYSTEMS.md?

benchmarks/apps/ramses/README.md
```diff
@@ -0,0 +1,43 @@
...
+reframe -c apps/ramses -r --performance-report
+```
+
+### Filtering the benchmarks
+
+By default all benchmarks will be run. You can run individual benchmarks with the
+[`--tag`](https://reframe-hpc.readthedocs.io/en/stable/manpage.html#cmdoption-0) option:
+
+* `weak` to run the weak scaling benchmarks
+* `strong` to run the weak scaling benchmarks
```
tkoskela, line 36: This looks like a typo
```suggestion
* `strong` to run the strong scaling benchmarks
```

benchmarks/spack/repo/packages/sphng/package.py
```diff
@@ -0,0 +1,69 @@
...
+            env["ENDIANFLAGLITTLE"] = "-convert little_endian"
+        elif self.compiler.name == "gcc":
+            fflags = (
+                "-m64 -mcmodel=large -O3 -I. -Wall -Wno-conversion -Wno-unused-dummy-argument "
+                "-Wno-maybe-uninitialized -Warray-temporaries"
+            )
+            env["DBLFLAG"] = "-fdefault-real-8 -fdefault-double-8"
+            env["DEBUGFLAG"] = "-g"
+            env["ENDIANFLAGBIG"] = "-fconvert=big-endian"
+            env["ENDIANFLAGLITTLE"] = "-fconvert=little-endian"
```
tkoskela, line 58: I played with building this using gcc and I ran into this error
```
densityiterate_gradh_MPI.F:2028:22:

 2028 |  29               IF (ifound.EQ.0) THEN
      |                      1
Error: Operands of comparison operator '.eq.' at (1) are LOGICAL(4)/INTEGER(4)
```
As far as I can tell, comparing integers and logicals in if statements is [not allowed by gfortran](https://gcc.gnu.org/onlinedocs/gfortran/Implicitly-convert-LOGICAL-and-INTEGER-values.html) so unless we want to touch the source code maybe just drop `gcc` from supported compilers?

I also had to add ` -fallow-argument-mismatch` to `fflags` to even get that far

TomMelt: Yeah, I agree. It requires code modification therefore we should just remove gcc support.

giordano (approved):

TomMelt:
Asif and I are just running final checks. We hope to be ready to merge later this week.

TomMelt:
Hi @tkoskela , it's ready to merge from our end.

SYSTEMS.md
```diff
@@ -71,6 +71,25 @@ reframe -c examples/sombrero -r --performance-report --system csd3-skylake:compu
...
+```
+reframe -c examples/sombrero -r --performance-report --system dial2:compute-node -J'--accout=<ACCOUNT>'
+```
+
+where `<ACCOUNT>` is the project you want to charge.
+
+This platform possesses the pbs/torque scheduler. In order to obtain and use the correct header option (`#PBS -l nodes=?:ppn=?`)
+the following additional flags are required (tested with sphng, trove, trove-pdsyev mpi/openmp hybrid apps):
+
+* `-S executable='-perhost $NUM_TASKS_PER_NODE <NAME-OF-EXECUTABLE>'`
```
giordano, line 90: Can this be achieved automatically with a custom launcher like https://github.com/ukri-excalibur/excalibur-tests/blob/615beef981c8bc4a1a243c5adedc365f4693929d/benchmarks/reframe_config.py#L5-L10

SYSTEMS.md
```diff
@@ -71,6 +71,25 @@ reframe -c examples/sombrero -r --performance-report --system csd3-skylake:compu
...
+* `-S env_vars=I_MPI_JOB_RESPECT_PROCESS_PLACEMENT:off`
```
giordano, line 91: If this is _always_ needed, you can set this environment variable in the partition configuration like https://github.com/ukri-excalibur/excalibur-tests/blob/615beef981c8bc4a1a243c5adedc365f4693929d/benchmarks/reframe_config.py#L48-L55

benchmarks/apps/trove/trove.py
```diff
@@ -0,0 +1,143 @@
...
+    def set_job_script_variables(self):
+
+        self.core_count_1_node = self.current_partition.processor.num_cpus // min(1, self.current_partition.processor.num_cpus_per_core)
+        self.num_tasks = self.num_mpi_tasks[self.param_value]
+        self.num_tasks_per_node = int(self.num_tasks/self.num_nodes_current_run[self.param_value])
+        self.num_cpus_per_task = int(self.core_count_1_node/self.num_tasks_per_node)
+
+        self.descr = ('Running Trove on ' +\
+                      str(self.num_nodes_current_run[self.param_value]) + ' node(s) with ' +\
+                      str(self.num_tasks_per_node) + ' tasks per node and ' +\
```
giordano, line 80: Style, I don't think you need the escape characters there?
```suggestion
        self.descr = ('Running Trove on ' +
                      str(self.num_nodes_current_run[self.param_value]) + ' node(s) with ' +
                      str(self.num_tasks_per_node) + ' tasks per node and ' +
```

benchmarks/apps/trove/trove.py
```diff
@@ -0,0 +1,143 @@
...
```
giordano, line 109:
```suggestion
        self.descr = ('Running Trove on ' +
                      str(self.num_nodes_current_run[self.param_value]) + ' node(s) with ' +
                      str(self.num_tasks_per_node) + ' tasks per node and ' +
```

benchmarks/apps/trove/trove.py
```diff
@@ -0,0 +1,143 @@
...
+    def set_job_script_variables(self):
+
+        self.core_count_1_node = self.current_partition.processor.num_cpus // min(1, self.current_partition.processor.num_cpus_per_core)
+        self.num_tasks = int(self.num_mpi_tasks[self.param_value])
+        self.num_tasks_per_node = int(self.num_tasks/self.num_nodes_current_run[self.param_value])
+        self.num_cpus_per_task = int(self.core_count_1_node/self.num_tasks_per_node)
+
+        self.descr = ('Running Trove on ' +\
+                      str(self.num_nodes_current_run[self.param_value]) + ' node(s) with ' +\
+                      str(self.num_tasks_per_node) + ' tasks per node and ' +\
```
giordano, line 138:
```suggestion
        self.descr = ('Running Trove on ' +
                      str(self.num_nodes_current_run[self.param_value]) + ' node(s) with ' +
                      str(self.num_tasks_per_node) + ' tasks per node and ' +
```

SYSTEMS.md
TomMelt: Do you know if there is a way to differentiate for intel and openMPI?

giordano: Ah, that's a bit annoying, for the linked COSMA7 partition I have two separate partitions, one with Intel MPI in the spack environment, and the other one with OpenMPI.  I wasn't particularly happy about duplicating the partition, but when launcher-specific configuration is needed that might be the easiest thing to do, for some definition of "easy".

SYSTEMS.md
TomMelt: In theory it doesn't need to be set for every job, just those that use intel MPI launcher. Which for our tests (dial tests) is all of them. But in theory it doesn't need to be. However, it should have no effect on openMPI jobs as the env variable should be ignored. So I can add it to the env_vars for dial2.5. Would you prefer that?

benchmarks/apps/trove/trove.py
TomMelt: Good spot. They are not required. I have removed them.

benchmarks/apps/sphng/sphng.py
```diff
@@ -0,0 +1,223 @@
...
+import reframe as rfm
+import reframe.utility.sanity as sn
+from benchmarks.modules.utils import SpackTest
+
+
+class SphngBase(SpackTest):
+
+    descr = 'Base class for Sphng'
+    valid_systems = ['*']
+    valid_prog_environs = ['*']
```
tkoskela, line 11:
```suggestion
    valid_prog_environs = ['default']
```
Could you enforce that these only run in the default programming environment? I'm going to submit another PR to clean up our environments, they don't do anything when we are building with spack, but just to avoid possible issues in the future. I tried to run these on Cosma8 and got flooded with benchmarks running in all environments.

benchmarks/apps/ramses/ramses.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @sanity_function
+    def validate_download(self):
+        return sn.assert_true(path.exists('data.tgz'))
+
+
+class RamsesMPI(SpackTest):
+
+    valid_systems = ['*']
+    valid_prog_environs = ['*']
```
tkoskela, line 24:
```suggestion
    valid_prog_environs = ['deault']
```
Related to comment on sphng

benchmarks/apps/trove/trove.py
```diff
@@ -0,0 +1,143 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+from benchmarks.modules.utils import SpackTest
+
+
+class Trove(SpackTest):
+
+    descr = 'Base class for Trove'
+    valid_systems = ['*']
+    valid_prog_environs = ['*']
```
tkoskela, line 10:
```suggestion
    valid_prog_environs = ['default']
```
Also related to comment on sphng :grinning:

benchmarks/apps/sphng/sphng.py
```diff
@@ -0,0 +1,223 @@
...
+    spack_spec = 'sphng@v1.0.0'
```
tkoskela, line 12:
```suggestion
    spack_spec = 'sphng@v1.0.0%intel'
```
Although we generally try to avoid it, I think here it would make sense to hard-code the intel compiler to the spec since it's the only supported compiler and the package recipe will fail with any other compiler.

benchmarks/apps/trove-pdsyev/pdsyev.py
```diff
@@ -0,0 +1,82 @@
...
+import reframe.utility.sanity as sn
+from benchmarks.modules.utils import SpackTest
+
+
+class PdsyevBase(SpackTest):
+
+    valid_systems = ['*']
+    valid_prog_environs = ['*']
+
+    spack_spec = 'trove-pdsyev@v1.0.0'
```
tkoskela, line 12:
```suggestion
    spack_spec = 'trove-pdsyev@v1.0.0%intel'
```

benchmarks/apps/trove/trove.py
```diff
@@ -0,0 +1,143 @@
...
+    spack_spec = 'trove@v1.0.0'
```
tkoskela, line 11:
```suggestion
    spack_spec = 'trove@v1.0.0%intel'
```

benchmarks/apps/ramses/ramses.py
```diff
@@ -0,0 +1,92 @@
...
+
+    spack_spec = 'ramses@v1.0.0'
```
tkoskela, line 26:
```suggestion
    spack_spec = 'ramses@v1.0.0%intel'
```

tkoskela (changes requested):

benchmarks/spack/repo/packages/sphng/package.py
```diff
@@ -0,0 +1,61 @@
...
+            fflags = (
+                "-O3 -mavx2 -mfma -mcmodel=medium -warn uninitialized -warn truncated_source "
+                "-warn interfaces -nogen-interfaces -DINCMPI"
+            )
+            env["DBLFLAG"] = "-r8"
+            env["DEBUGFLAG"] = "-check all -traceback -g -fpe0 -fp-stack-check -heap-arrays -O0"
+            env["ENDIANFLAGBIG"] = "-convert big_endian"
+            env["ENDIANFLAGLITTLE"] = "-convert little_endian"
+        else:
+            msg = "The compiler [{self.compiler.name}] is not supported yet."
```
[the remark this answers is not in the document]
TomMelt: I'll fix this. It's missing the "f" infront of the quotes.

benchmarks/apps/sphng/sphng.py
tkoskela: Technically I suppose the `oneapi` compiler would also work, but I would expect to always have an `intel` compiler available if `oneapi` is installed.

benchmarks/apps/sphng/sphng.py
TomMelt: done

benchmarks/spack/repo/packages/sphng/package.py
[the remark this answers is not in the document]
TomMelt: done for all DiRAC apps

benchmarks/apps/sphng/sphng.py
TomMelt: OK I will add `intel`. the user will have to overwrite the recipe if `intel` is not available.

SYSTEMS.md
TomMelt: @giordano So just to be clear, would you like me to make two separate partitions for Dial2.5? or keep the code as is?

SYSTEMS.md
TomMelt: @giordano also do you want me to default add the env_var for all jobs or leave it as is?

benchmarks/apps/sphng/sphng.py
tkoskela: Thanks, I think that's better than user always having to overwrite the recipe if `intel` is not the default!

SYSTEMS.md
TomMelt: Actually I had another thought. This will still require manual work on dial2.5 because users will need to overwrite the running partition anyway. I'd be inclined to leave this as is.

giordano: Uhm, ok.

SYSTEMS.md
giordano: If `I_MPI_JOB_RESPECT_PROCESS_PLACEMENT` is needed for all Intel MPI jobs and it's ignored by other MPI implementations, I think it'd still be easier to set it in the partition settings, yes.

SYSTEMS.md
TomMelt: sorry I appreciate it's not ideal but we will be getting rid of dial2.5 soon (possible this year) so it doesn't feel like the best use of time.

SYSTEMS.md
asifsamiarain: Thanks @giordano, this particular flag is needed for mpi/opnmp hybrid apps only. 

I am working on a solution and will commit shortly for final review.

tkoskela (approved):
I got this running on csd3, looks good to me now. Having some issues with cosma7/8 but I suspect that has to do with our spack environment.

tkoskela (commented):
I did a bit more testing, and noticed that `sphng` is asking for up to 128 tasks per node in the `single_node_ifile` test, and this can fail on systems with smaller nodes. Could there be an assertion that checks how many cores per node are available?

giordano:
Number of (logical) CPUs per node can be accessed with `self.current_partition.processor.num_cpus`, when doing scaling benchmarks you may want to do parametrise the number of tasks by this figure, instead of hardcoding number like 128.

TomMelt:
> Number of (logical) CPUs per node can be accessed with `self.current_partition.processor.num_cpus`, when doing scaling benchmarks you may want to do parametrise the number of tasks by this figure, instead of hardcoding number like 128.

@tkoskela @giordano My initial thought was to write something like this:
```python
class Sphng_Single_Node_ifile(SphngBase_ifile):

    num_cpus = self.current_partition.processor.num_cpus
    max_power_of_2 = int(np.log2(num_cpus)) + 1
    mpi_tasks = parameter(2**i for i in range(0, max_power_of_2))
```

But AFAIK this will not work for `parameters`. The `self.current_partition.processor.num_cpus` variable is only set in the `setup` stage of reframe tests. So I don't think we can use `self.current_partition.processor.num_cpus` to set the upper limit of the test and `parameter` has to be set in the test class body. If you can think of another way please let me know.

giordano:
> The self.current_partition.processor.num_cpus variable is only set in the setup stage of reframe tests. So I don't think we can use self.current_partition.processor.num_cpus to set the upper limit of the test and parameter has to be set in the test class body. If you can think of another way please let me know.

I asked in the ReFrame slack and was told to look at https://github.com/reframe-hpc/reframe/blob/5e5e94c3655b58704f26725871ad42d4bd51d8e2/tutorials/cscs-webinar-2022/tests/stream9.py#L63-L75

TomMelt:
> > The self.current_partition.processor.num_cpus variable is only set in the setup stage of reframe tests. So I don't think we can use self.current_partition.processor.num_cpus to set the upper limit of the test and parameter has to be set in the test class body. If you can think of another way please let me know.
> 
> I asked in the ReFrame slack and was told to look at https://github.com/reframe-hpc/reframe/blob/5e5e94c3655b58704f26725871ad42d4bd51d8e2/tutorials/cscs-webinar-2022/tests/stream9.py#L63-L75

Thanks @giordano but this still doesn't solve our problem. The issue is that you could in theory have 2 partitions with different `processor.num_cpus` e.g., `isambard-cascadelake`. In this case they happen to have the same number of cpus but I guess we can't count on that. So then we would be left with the same issue that the test fails if it runs on the partition with a smaller number of cpus. We really need the *current* `processor.num_cpus` however we cannot get it until the `setup` phase.

Do you know if there is there a way to graciously skip a test with some appropriate warning? rather than straight up fail?

Otherwise, the only way around it would be to remove all of the scaling tests and the user would have to bash script the tests for each system they want to run on.

TomMelt:
@giordano , nvm something like this will do the job for now. Although there is the edge case I mentioned for systems with multiple partitions.

```python
    procs_per_part = [p.processor.num_cores for p in rt.runtime().system.partitions]
    max_tasks = np.max(procs_per_part)
    max_power_of_2 = int(np.log2(max_tasks)) + 1 
    mpi_tasks = parameter(2**i for i in range(0, max_power_of_2))
```

benchmarks/apps/sphng/sphng.py
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+
+class Sphng_Single_Node_ifile(SphngBase_ifile):
+
+    mpi_tasks = parameter(threads_per_part(), fmt=lambda x: x[1])
+
+    @run_after('init')
+    def setup_thread_config(self):
+        self.valid_systems = [self.mpi_tasks[0]]
```
TomMelt, line 89: @tkoskela @giordano I have managed to use the feature they discussed to get mpi_tasks to be generated correctly for all partitions. To check I just made a dummy partition in the reframe_config with only 4 cpus and it generates the correct number of tests

TomMelt: So I think we are finally ready to merge now :sweat_smile:

tkoskela: This is probably a stupid question, but why does `self.mpi_tasks` go into `self.valid_systems`?

TomMelt: It's not a stupid question :wink: 

I took this directly from [the example](https://github.com/reframe-hpc/reframe/blob/5e5e94c3655b58704f26725871ad42d4bd51d8e2/tutorials/cscs-webinar-2022/tests/stream9.py#L63-L75). 

`parameter(threads_per_part(), fmt=lambda x: x[1])` returns a tuple containing the system partition name and the number of mpi tasks. 

The issue as I understand is that the performance variable gets it's name from the parameter variable. So when test names and parameters are printed it will read as e.g., `mpi_tasks=4` due to the `fmt=lambda x: x[1]` part. I agree that the name is misleading but this is so that it fits nicely into the rest of reframe.

TomMelt:
Hi @giordano , @tkoskela , we are ready to merge now -- if that's ok with you.

tkoskela:
Ok, with me :+1:

PR #134, opened 2023-04-24
[Isambard] Move `memory` to its own resource for Nvidia Ampere partition

giordano:
Also, add items to `extra_resources` dictionary, instead of resetting it, which would override a setting on the command line.

### DIFF
--- a/SYSTEMS.md
+++ b/SYSTEMS.md
@@ -97,6 +97,12 @@ To run compilation on the compute node, you have to set the attribute [`build_lo
 reframe -c examples/sombrero -r --performance-report --system isambard-cascadelake:compute-node -S build_locally=false
 ```
 
+### Phase3 partition
+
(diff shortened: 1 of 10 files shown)


SYSTEMS.md
```diff
@@ -97,6 +97,12 @@ To run compilation on the compute node, you have to set the attribute [`build_lo
 reframe -c examples/sombrero -r --performance-report --system isambard-cascadelake:compute-node -S build_locally=false
 ```
 
+### Phase3 partition
+
+On the nodes with Nvidia Ampere GPUs, memory usage is restricted to 4 GiB by default.
+To request more memory you [have to specify the resource](https://gw4-isambard.github.io/docs/user-guide/PHASE3.html#nvidia-gpu) `mem=...G`, for example by setting the `memory` key of `extra_resources` to the amount of memory you require.
+This can be done on the command line of ReFrame [...]
```
giordano, line 104: This requires https://github.com/reframe-hpc/reframe/issues/2868

PR #137, opened 2023-05-04
Set `self.build_system.specs` in base `SpackTest` class

giordano:
This reduces the boilerplate code needed to set up a new test.


tkoskela (commented):
Nice! I like less code

Did you miss [The sombrero example](https://github.com/ukri-excalibur/excalibur-tests/blob/469933627ac346b8b3079c2f874c0dae16a1f151/benchmarks/examples/sombrero/sombrero.py#L107-L108)?

PR #139, opened 2023-05-04
Set `self.extra_resources` directly in base `SpackTest` class

giordano:
In the same vein as #137, further reduction of boilerplate code.

I need to double check this on Myriad, which is currently down, but I'm mostly confident this works as expected.


tkoskela:
Tried it and it works for me!

PR #147, opened 2023-05-10
Add some minimal documention for how to write GPU benchmarks

giordano:
CC: @kaanolgu.


kaanolgu:
Tested the steps and it worked. They were easy to follow thanks!

giordano:
Awesome, thanks for trying this out!

Issue #48, opened 2022-10-17
Matrix free hpcg benchmark


tkoskela:

dcaseGH:
A matrix free HPCG has been saved and tagged as version one here:

https://github.com/NCAS-CMS/hpcg_27ptStencil
this has been used in Reframe and Spack - DHC will contact Mose about this. They should work for any version of HPCG.

The next steps are to modify HPCG to solve an 'LFRic problem'.

tkoskela:
After we merge #85 add Spack recipe for this benchmark and create `hpcg` in `apps`

dcaseGH:
There is a branch of the 27pt stencil repository on the NCAS-CMS github for the LFRic problem
https://github.com/NCAS-CMS/hpcg_27ptStencil/tree/lfric_stencil_data

There is a file (dinodump.dat) which is read in, and information attached to the geometry objects etc with data and 'grid' dimensions (check its there - it's from a 24*24*6*30 grid).
The GS preconditioning (with mutligrid) is turned off, although a stencil version of GS could be implemented if wanted.

The HPCG runs tests such as a.A.b - b.A.a to confirm A is symmetric - this fails atm - must convince myself whether this should pass or not and why.

dcaseGH:
I have included 2 tests in the hpcg_benchmarks branch (commit [3bb31c7](https://github.com/ukri-excalibur/excalibur-tests/commit/3bb31c70fb9fc607d28bac54d0f8be88b5e8e76b)), in the apps/hpcg section, with an appropriate spack-environments/repo/packages/hpcg_excalibur/package.py
One, version 1.0 gets a copy of the code to run the 27 pt stencil
The other, version hpcg_lfric gets a copy of a code with a different tag, and makes that for lfric

However, the second part doesn't appear to be installing this version properly, and at the run stage it appears that version 1.0 is run twice, with version hpcg_lfric not run at all. I need to get my wires uncrossed wrt spack next week.

dcaseGH:
Error was giving `version@version` as spack spec, not `spec@version` - fixed at [4dac6f4](https://github.com/ukri-excalibur/excalibur-tests/commit/4dac6f4e06e8c8dcd1869b1f13fa68960d3d256c)

Need vanilla HPCG; refactoring and cleaning reframe/spack; updating hpcg code to have verified outputs etc

tkoskela:
Closing this issue as completed by #157

PR #157, opened 2023-05-22
initial branch - attempt to get this to work on isambard cascadelake …

dcaseGH:
#48 
@giordano  - I'm going to use this pull request to try to merge the hpcg benchmarks. I've made a new branch from the head of main, but this has quite a few changes and I'm not sure about the status of some of the things. I'll record what I'm doing here - but note that this didn't run (as will be explained). If you get some time to let me know what I should be doing to run on cascade lake at Isambard, that'd be really helpful.

I'm going to talk out loud here, so that I remember what I've done, and so that anyone can tell me where I'm going wrong if they read this.

source ./excalibur_hpcg/bin/activate
module load python37 (as you guys have discussed on slack - this is needed for pip)
pip install .

export SPACK_ROOT="/home/mox-dcase/spack"
source "${SPACK_ROOT}/share/spack/setup-env.sh"
export RFM_CONFIG_FILES="${PWD}/benchmarks/reframe_config.py"

reframe -c benchmarks/apps/hpcg -r --keep-stage-files -n HPCG_Original --system isambard-macs:cascadelake
(This is the command that I want to run - HPCG_Original is one of the three variants of the hpcg app. This will fail because benchmarks/spack/isambard-macs/packages.yaml has references to 'open' in perl and crypt in python, which can't be found).

reframe version is 4.2.0
spack version is 0.19.0.dev0

So the status after the first attempt (from this new branch) is that there is an issue resolving the cascade lake environment for spack, and this needs prevents it from moving on to attempt to build the apps.

### DIFF
--- a/benchmarks/apps/hpcg/hpcg.py
+++ b/benchmarks/apps/hpcg/hpcg.py
@@ -35,7 +35,9 @@ class HPCGBenchmark(SpackTest):
     executable_opts = []
     # Time limit of the job, automatically set in the job script.
     time_limit = '3m'
-
+    # hpcg.dat sets size of grid
+    prerun_cmds.append('cp "$(dirname $(which xhpcg))/hpcg.dat" .')
+    
     reference = {
         # flops? Gflops/s ?
         'archer2': {
@@ -92,26 +94,21 @@ class HPCGBenchmark(SpackTest):
                              self.current_partition.processor.num_cpus // 
                              self.num_cpus_per_task) 
 
-    @run_after('compile')
-    def copy_input_file(self):
-        # during install data files will be copied to the spack env bin - put them in stagedir to run with
-        # hpcg.dat sets grid size for the run - 104x104x104
-        self.prerun_cmds = ["python -c 'import shutil;import os;shutil.copyfile(shutil.which(\"xhpcg\").replace(\"xhpcg\",\"hpcg.dat\"), os.getcwd()+\"/hpcg.dat\")'"]
         
 @rfm.simple_test
 class HPCG_Stencil(HPCGBenchmark):
     spack_spec = 'hpcg_excalibur@hpcg_stencil'
+    tags = {"27pt_stencil"}
     
 @rfm.simple_test
 class HPCG_Original(HPCGBenchmark):
     spack_spec = 'hpcg_excalibur@hpcg_original'
+    tags = {"27pt_stencil"}
     
 @rfm.simple_test
 class HPCG_LFRic(HPCGBenchmark):
     # As above but with the LFRic style stencil
     spack_spec = 'hpcg_excalibur@hpcg_lfric'
-
-    @run_before('run')
-    def copy_lfric_input_file(self):
-        # lfric app requires extra data - dinodump.dat
-        self.prerun_cmds += ["python -c 'import shutil;import os;shutil.copyfile(shutil.which(\"xhpcg\").replace(\"xhpcg\",\"dinodump.dat\"), os.getcwd()+\"/dinodump.dat\")'"]
+    tags = {"lfric"}
+    # lfric app requires extra data - dinodump.dat
+    prerun_cmds.append('cp "$(dirname $(which xhpcg))/dinodump.dat" .')
--- a/benchmarks/apps/hpcg/hpcg.py
+++ b/benchmarks/apps/hpcg/hpcg.py
@@ -45,7 +45,7 @@ class HPCGBenchmark(SpackTest):
     @run_after('setup')
     def setup_variables(self):
         # Strictly HPCG is only intended to run for 1 OMP thread, except in original version
-        self.variables['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
+        self.env_vars['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
 
     @run_before('compile')
     def setup_build_system(self):
@@ -78,7 +78,6 @@ class HPCGBenchmark(SpackTest):
         self.set_var_default('num_tasks_per_node', 
                              self.current_partition.processor.num_cpus // 
                              self.num_cpus_per_task) 
-        self.env_vars['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
         
         
 @rfm.simple_test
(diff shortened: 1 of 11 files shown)


benchmarks/apps/hpcg/hpcg.py
```diff
@@ -0,0 +1,100 @@
...
+
+import os.path as path
+import sys
+import reframe as rfm
+import reframe.utility.sanity as sn
+import glob
+
+# Add top-level directory to `sys.path` so we can easily import extra modules
+# from any directory.
+sys.path.append(path.join(path.dirname(__file__), '..', '..'))
```
giordano, line 14: This isn't needed anymore
```suggestion
```

benchmarks/apps/hpcg/hpcg.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @run_after('setup')
+    def setup_variables(self):
+        # Strictly HPCG is only intended to run for 1 OMP thread, except in original version
+        self.variables['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
+
+    @run_before('compile')
+    def setup_build_system(self):
+        self.build_system.specs = [self.spack_spec]
+
```
giordano, line 60: Not needed anymore thanks to #137
```suggestion
```

benchmarks/apps/hpcg/hpcg.py
```diff
@@ -0,0 +1,100 @@
...
+    @run_before('sanity')
+    def set_sanity_patterns(self):
+        # Should probably check that it's a valid run
+        self.sanity_patterns = sn.assert_found(r'', self.stdout)
```
giordano, line 64: Maybe this can check something more useful than nothing?  For example, is there anything written to stdout/stderr that could be useful to match?

giordano (commented):
First round of comments

benchmarks/spack/repo/packages/hpcg_excalibur/package.py
```diff
@@ -0,0 +1,57 @@
...
+#        ''' Makefile has no install, but spack seems to want one for this package, so do build instead of install here (or inherit something else??) '''
+#        return ['arch=Linux_MPI', "CXXFLAGS='-DHPCG_NO_OPENMP'"]#, 'install']
+    
+    def install(self, spec, prefix):
+        ''' Makefile has no install '''
+        mkdirp(prefix.bin)
+        # copy dinodump.dat to the bin dir (same place as bin as easy to find with reframe)
+        if self.spec.satisfies("@hpcg_lfric"):
+            import shutil
+            shutil.copyfile('dinodump.dat', prefix.bin + '/dinodump.dat')
```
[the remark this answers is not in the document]
dcaseGH: There are two text files that are included in the LFRic version of the app. As they're text, they're just in the source. A better way of implementing this could be found later, no doubt.

benchmarks/apps/hpcg/hpcg.py
```diff
@@ -0,0 +1,100 @@
...
+            'flops': sn.extractsingle(
+                r'VALID with a GFLOP/s rating of=(\S+)',
+                self.output_data, 1, float),
+        }
+
+@rfm.simple_test
+class HPCG_Stencil(HPCGBenchmark):
+    spack_spec = 'hpcg_excalibur'        
+    num_tasks = 40
+    num_tasks_per_node = 40
```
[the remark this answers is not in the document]
dcaseGH: I stuck this in and it seemed to run on on 1 core??? Will investigate

benchmarks/apps/hpcg/hpcg.py
dcaseGH: There is something in the self.perf_patterns which looks for the word 'Valid ....' in the output file. This could be used again (check only one possible_outfiles too)

dcaseGH:
I've made some changes with respect to @giordano 's suggestions, and also to keep a standard grid for the stencil to original comparison.
If you want to run this you need spack vn 0.21.0.dev0 (not 0.19 as in the header)
and argument to reframe:  -S build_locally=false (Thanks Mose)

Only tested on cascade lake @ isambard , and only openmpi at this stage.

benchmarks/apps/hpcg/hpcg.py
```diff
@@ -0,0 +1,117 @@
...
+    spack_spec = 'hpcg_excalibur@hpcg_stencil'
+    
+@rfm.simple_test
+class HPCG_Original(HPCGBenchmark):
+    spack_spec = 'hpcg_excalibur@hpcg_original'
+    
+@rfm.simple_test
+class HPCG_LFRic(HPCGBenchmark):
+    # As above but with the LFRic style stencil
+    spack_spec = 'hpcg_excalibur@hpcg_lfric'
```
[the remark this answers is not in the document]
dcaseGH: Sure. I've added the same tag for 2 of them, as you may want to run them together. If you want to just call them individually by name there is of course functionality for that in reframe.

benchmarks/apps/hpcg/hpcg.py
```diff
@@ -0,0 +1,110 @@
...
+        
+@rfm.simple_test
+class HPCG_Stencil(HPCGBenchmark):
+    spack_spec = 'hpcg_excalibur@hpcg_stencil'
+    tags = {"27pt_stencil"}
+    
+@rfm.simple_test
+class HPCG_Original(HPCGBenchmark):
+    spack_spec = 'hpcg_excalibur@hpcg_original'
+    tags = {"27pt_stencil"}
```
giordano, line 102: Is it expected that these benchmarks share the same tag name?

dcaseGH: Yes. I thought that if you wanted to run a test on its own, you would give the name argument to reframe (-n/--name). But if you wanted to run a collection, for example all those who are based on a 27 pt stencil problem, you could use the tag. We only have 3 apps in the collection at the moment, but will hopefully have more later so this will be a more important issue.
This is just what seemed good - if we have a different convention then I will follow that.

giordano (commented):
Looks overall good to me after including a README file.  We can improve the benchmark and/or the Spack recipe in the future.

dcaseGH:
Ok - I've added the readme - thanks. I'll of course return to all of this as we decide what's needed following our meeting next month. See you then.

giordano:
Thanks!

PR #161, opened 2023-06-02
Install pandas and pytest as dependencies of excalibur-tests

tkoskela:
Following up from my [review comments](https://github.com/ukri-excalibur/excalibur-tests/pull/121#discussion_r1205538810) I thought we should be able to `pip install` everything we need in one go.

- It looks like setting the `--target` flag to `pip` was breaking `pytest` on the GHA runners, the error is in [this failed action](https://github.com/tkoskela/excalibur-tests/actions/runs/5153671619/jobs/9281190331)
- I first thought of using a virtual environment but again [struggled with setting the path correctly](https://github.com/tkoskela/excalibur-tests/actions/runs/5155188057/jobs/9284547982)
- In the end, it works by just installing into the default python path which now seems short enough for spack.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,12 +33,12 @@ jobs:
         with:
           python-version: '3.10.6'
           cache: 'pip'
-      - name: Install Reframe
+          
+      - name: Install Excalibur-tests
         shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install . --target ${{runner.workspace}}/pip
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -33,12 +33,12 @@ jobs:
         with:
           python-version: '3.10.6'
           cache: 'pip'
-      - name: Install Reframe
+
```
pineapple-cat, line 36:
```suggestion   
```
Small bit of unnecessary/inconsistent whitespace.

pineapple-cat (approved):
Right, I forgot to add a pip install for the entire project in `ci.yml`! These changes look good to me.

Issue #104, opened 2023-03-08
Create high-level script to run postprocessing


ilectra:
This should accept some input either in command line or a config yaml/json file, and based on that it should perform the required type of analysis (see use cases in https://github.com/ukri-excalibur/excalibur-tests/issues/70#issue-1522882139).
The steps of this analysis should be something like
- read desired data from some file(s) in a path and store in a pandas dataframe (possibly collate results of various files)
- call the scripts to produce required output (plots and/or tables). The input to those scripts will be a subset of the dataframe, but its exact form cannot be decided yet, until we have some plotting scripts.
- produce info output with the dataframe column names, or a debug output of a dump of the whole dataframe

PR #166, opened 2023-06-09
Add high-level control script for post-processing

pineapple-cat:
Addresses #104.

- [x] Small update to display name parsing (unrelated QOL update)
- [x] Command line argument parsing
- [x] Get perflog information to pass to plotting scripts
- [x] Display the relevant dataframe columns used
- [x] Unit tests for dataframe collation


post-processing/post_processing.py
```diff
@@ -1,18 +1,89 @@
...
+        # identify a relevant subset of the dataframe to pass on
+        # TODO
+        columns = ["test_name", "system", "partition", "environ"]
+        for fom in figures_of_merit:
+            value = fom + "_value"
+            unit = fom + "_unit"
+            if (value in df.columns) & (unit in df.columns):
+                columns.extend([value, unit])
+            else:
+                if self.debug:
```
pineapple-cat, line 59: Should be visible even without debug flag (print all available columns and throw an error).

post-processing/post_processing.py
```diff
@@ -1,18 +1,89 @@
...
+            print("Found log files:")
+            for log in log_files:
+                print("-", log)
+
+        df = pd.DataFrame()
+        # put all perflog information in one dataframe
+        for file in log_files:
+            try:
+                temp = read_perflog(file)
+                df = pd.concat([df, temp], ignore_index=True)
```
pineapple-cat, line 42: Double-check how this works with introducing new columns.

post-processing/post_processing.py
```diff
@@ -1,18 +1,89 @@
...
+            except KeyError as e:
+                if self.debug:
+                    print("Discarding %s:" %os.path.basename(file), type(e).__name__ + ":", e.args[0], e.args[1])
+
+        if df.empty:
+            raise FileNotFoundError(errno.ENOENT, "Could not find a valid perflog in path", path)
+
```
pineapple-cat, line 49: Add a loop for optional filtering of rows (e.g. by date, by system, etc.). Need to have this as an input to the run function (could mean some columns that were filtered by are no longer necessary).

If this is empty (e.g. no rows passed the filters), throw an error.

pineapple-cat:
Implemented suggestions:

- [ ] Change `run_post_processing` input to get a path to a config file containing:
   a.) user-selected columns
   b.) filter conditions for row selection
- [ ] Make invalid columns throw an error + check there are at least 3 columns specified:
   a.) figure of merit value
   b.) figure of merit unit
   c.) something to plot the figure of merit against
- [ ] Add optional row filtering function + throw an error if this results in an empty DataFrame
- [ ] Add test for concatenating DataFrames with different columns

post-processing/post_processing.py
```diff
@@ -43,33 +49,107 @@ def run_post_processing(self, path, figures_of_merit):
...
+        # evaluate expression and extract dataframe mask
+        if value is None:
+            if operator == op.eq:
+                mask = df[column].isnull()
+            else:
+                mask = df[column].notnull()
+        else:
+            try:
+                # dataframe column is interpreted as the same type as the supplied value
+                mask = operator(df[column].astype(type(value)), value)
```
ilectra, line 132: There's a slight issue here, if the type of values of the column doesn't agree with the type of the filter value from the YAML (eg. int vs float). But this is part of a larger issue, that will be resolved separately, see #176

post-processing/post_processing_config.yaml
```diff
@@ -0,0 +1,23 @@
+# Plot title
```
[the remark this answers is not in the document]
pineapple-cat: Opened #178 for this

post-processing/post_processing.py
```diff
@@ -1,41 +1,179 @@
...
+
+        invalid_columns = []
+        # check for invalid columns
+        for col in columns:
+            if col not in df.columns:
+                invalid_columns.append(col)
+        if invalid_columns:
+            raise KeyError("Could not find columns", invalid_columns)
+
+        filters = config["filters"]
```
ilectra, line 70: A check just like the one you have above, to make sure the needed columns for the filters are included in the dataframe, has to be added here. Same for the datasets.
Note that this check has to be here and not in the `read_config`, as is compares to the df.

post-processing/post_processing.py
```diff
@@ -1,41 +1,179 @@
...
+        mask = pd.Series(df.index.notnull())
+        # filter rows
+        if filters:
+            mask = reduce(op.and_, (self.row_filter(f, df) for f in filters))
+        if df[mask].empty:
+            raise pd.errors.EmptyDataError("Filtered dataframe is empty", df[mask].index)
+
```
ilectra, line 77: If datasets have been given in the config
- their columns should be added to the `columns` here
- the number of rows after the filtering should be = (number of x points)x(number of datasets). If no datasets are given, then (number of rows)=(number of x points).

post-processing/post_processing_config.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# Optional setting to display several plots in the same graph
+# Dataset format: [column_name, value]
+datasets: [["system", "default"]]
+
+# Axes must have a value specified with a column name
+# Units can either be specified with a column name
+# or a custom label (including null (i.e. None) if units are N/A)
+
+x_axis:
```
ilectra, line 23: More checks to be added to `read_config`: the value and units for both axis should be the same as the `columns` given above.

post-processing/post_processing.py
ilectra: I should say "number of _unique_ x points". Maybe read that column into a set to get this number?

post-processing/post_processing_config.yaml
pineapple-cat: Depending on what we need additional columns for, we could potentially remove the columns from config entirely and just internally compile a necessary columns list based on axis values, units, and datasets.

ilectra (approved):

PR #174, opened 2023-06-26
[isambard-macs] Comment out unusable `intel-mpi` external packages

### LINKED ISSUE #171: Intel mpi configuration on Isambard
`intel-mpi` in our current spack environment is not correctly configured. It fails to find `mpi.h` when building and `mpirun` when running. It seems to get selected as the default mpi implementation. We should either fix it, or make openmpi the default.

giordano:
I tried different combinations of `prefix` and `modules`, but could never make spack find mpi correctly.  Fix #171.


tkoskela (approved):

giordano:
For the record, I think I tried every bit of the path from `/cm/shared/apps/intel` to `/cm/shared/apps/intel/compilers_and_libraries/2020.4.304/linux/mpi/intel64`, always to no avail.  I think spack just doesn't like the layout of Intel MPI on this system and doesn't find it.

PR #175, opened 2023-06-26
Fix post-processing CI test for ReFrame v4.3.0

pineapple-cat:


giordano:
I think we want to explicitly require reframe 4.3.0 in https://github.com/ukri-excalibur/excalibur-tests/blob/d236363f7f58a8b0b2cf1e5619f356eb5788ae75/pyproject.toml#L26 ?

PR #179, opened 2023-07-05
Fix sample code in `SYSTEMS.md` file

giordano:
These were still referring to the old paths before changing the layout of the repository.


tkoskela (approved):
Well spotted!

PR #184, opened 2023-07-12
Remove reference to installing ReFrame manually

giordano:
This is no longer needed and this line can create confusion.


tkoskela (approved):
I suppose you are right, if someone really needs to install it manually they can probably figure out how to do it. Not worth confusing the average user over it.

PR #185, opened 2023-07-13
Remove libffi because it's not there anymore

tkoskela:


giordano (approved):

giordano:
I think lots of the packages previously in `/usr` aren't there anymore since
```console
giordano@ln04:~> ls /usr/lib/lib*.so
/usr/lib/libAtpDispatch.so  /usr/lib/libs3.so
```

tkoskela:
This is now implemented in https://github.com/ukri-excalibur/excalibur-tests/pull/220

Issue #107, opened 2023-03-08
Generic scripts for creating plots/tables of benchmark results


ilectra:
Or "use case 1" in https://github.com/ukri-excalibur/excalibur-tests/issues/70#issue-1522882139 , duplicated here:

> Single benchmark app run with various parameters (eg. [castep](https://github.com/ukri-excalibur/excalibur-tests/blob/main/apps/legacy/castep/castep.ipynb)). Each app can have different levels of things to create outputs/visualisations for, eg. different FOMs vs # nodes for different systems or for different input parameters, or both. The generic script should take the name of those FOMs/other parameters to create plots for, and fish them out of the DF.

Create github-pages with the output, re-using as much as possible the stuff in https://github.com/t-young31/excalibur-tests and https://github.com/ukri-excalibur/excalibur-tests/issues/18

PR #188, opened 2023-07-14
Add generic plot script for post-processing

pineapple-cat:
Addresses #107.

- [x] Rename `datasets` to `series` (unrelated change based on feedback requests)
- [x] Create simple bar chart to plot numeric figure of merit (y-axis) vs something else (x-axis)
- [x] Add optional series groupings to bar chart x-axis
- [x] Add title, labels, and tooltips to charts
- [x] Figure out colour mapping + legend
- [x] Add some documentation and make a tutorial for post-processing

### DIFF
--- a/post-processing/post_processing.py
+++ b/post-processing/post_processing.py
@@ -135,6 +135,16 @@ class PostProcessing:
         # get column names of axes
         x_column = x_axis.get("value")
         y_column = y_axis.get("value")
+        # get units
+        x_units = df[x_axis["units"]["column"]][0] if x_axis.get("units").get("column") \
+                  else x_axis.get("units").get("custom")
+        y_units = df[y_axis["units"]["column"]][0] if y_axis.get("units").get("column") \
+                  else y_axis.get("units").get("custom")
+        # determine axis labels
+        x_label = "{0}{1}".format(x_column.replace("_", " ").title(),
+                                  " ({0})".format(x_units) if x_units else "")
+        y_label = "{0}{1}".format(y_column.replace("_", " ").title(),
+                                  " ({0})".format(y_units) if y_units else "")
 
         # find x-axis groups (series columns)
         groups = [x_column]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "reframe-hpc >= 4.3.0, < 5.0.0",
     "matplotlib >= 3.0.0",
     "pandas >= 2.0.1",
+    "bokeh >= 3.2.0",
```
pineapple-cat, line 29: Using latest bokeh version for now, but we can likely find an older version that works if needed.

post-processing/post_processing_tutorial.md
```diff
@@ -0,0 +1,25 @@
...
+
+There are three main post-processing components:
+- **`Perflog parsing:`**
+  - Data from benchmark performance logs are stored in a pandas DataFrame.
+- **`Data filtering:`**
+  - If more than one perflog is used for plotting, DataFrames from individual perflogs are concatenated together into one DataFrame.
+  - The DataFrame is then filtered, keeping only relevant rows and columns.
+- **`Plotting:`**
+  - A filtered DataFrame is passed to a plotting script, which produces a graph and embeds it in a simple HTML file. (`TODO: Embed graphs in GitHub Pages.`)
+  - Users may run the plotting script to generate a generic bar chart. Graph settings should be specified in a configuration YAML file. (`TODO: Add scaling and regression plots.`)
```
pineapple-cat, line 15: Should to-dos be kept or removed from the tutorial? It seems pertinent to inform users of missing features and this is easier than reworking more of the tutorial later, but it would look a bit neater to not have them there.

post-processing/post_processing.py
```diff
@@ -111,6 +120,61 @@ def run_post_processing(self, log_path, config):
...
+        x_column = x_axis.get("value")
+        y_column = y_axis.get("value")
+        # get units
+        x_units = df[x_axis["units"]["column"]][0] if x_axis.get("units").get("column") \
+                  else x_axis.get("units").get("custom")
+        y_units = df[y_axis["units"]["column"]][0] if y_axis.get("units").get("column") \
+                  else y_axis.get("units").get("custom")
+        # determine axis labels
+        x_label = "{0}{1}".format(x_column.replace("_", " ").title(),
+                                  " ({0})".format(x_units) if x_units else "")
```
pineapple-cat, line 145: Should other group column names also be added to the x-axis label?

post-processing/post_processing_tutorial.md
ilectra: I'll put them in a separate section at the end. I'll also re-name this into a README, and pick and choose for the actual tutorial. But it's very close! :)

post-processing/post_processing.py
ilectra: Yes, but I'll move this to a new issue, since we have to fix the 2-level grouping anyway.

ilectra: See #198

Issue #206, opened 2023-09-05
Make postprocessing dependencies optional


tkoskela:
It seems that we have implicitly raised our required python version to 3.9 by requiring bokeh-3.2.0. Do we need that version or can we live with an older one?

tkoskela:
Luckily Archer2 has python 3.9 as default :sweat_smile:

tkoskela:
We should make postprocessing an optional dependency so that the framework can be installed without it, similar to the tests. See

https://github.com/ukri-excalibur/excalibur-tests/blob/2d86770998241e45c63a0548d1fcf225c620e371/pyproject.toml#L32-L35

Issue #201, opened 2023-08-30
Problem parsing display_name field


tkoskela:
If a perflog contains both parametrized and non-parametrized runs, the `display_name` field is not parsed properly if the first entry is not parametrized.

Issue #199, opened 2023-08-25
Parse the extra_resources and env_vars perflog fields


ilectra:
We need to unpack those dictionaries into individual df columns, similar (but easier, hopefully!) to fishing out the params from the `display_name` fields.

Issue #176, opened 2023-06-27
Fix DataFrame types for post-processing


pineapple-cat:
When reading perflogs into a DataFrame, pandas automatically converts all columns to either an object or a string. There needs to be a fix for this so the high-level script filters and plotting scripts can all work properly.

One possible way to do this is to include the units of all selected columns in the YAML config. See also `pd.to_numeric(errors='ignore')`.

As part of this, also make sure that the timestamp column is properly taken care of by using `datetime` (eg. comparison operators in filters based on this column should work seamlessly by using `datetime`)

PR #207, opened 2023-09-15
Add proper DataFrame types for post-processing (+ other misc fixes)

pineapple-cat:
Addresses #176.

- [x] Add proper numerical data type usage
- [x] Introduce datetime type to ensure correct filtering of `job_completion_time`

Other fixes:
- [x] #206
- [x] #165
- [x] #201
- [x] #199

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
         shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install .[test,post-processing]
+          pip install .[test]
       - name: Install Spack
         shell: bash
         working-directory: ${{runner.workspace}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
         shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install .[test]
+          pip install .[test,post-processing]
       - name: Install Spack
         shell: bash
         working-directory: ${{runner.workspace}}
--- a/post-processing/post_processing.py
+++ b/post-processing/post_processing.py
@@ -146,20 +146,9 @@ class PostProcessing:
                 series_filters: list, x-axis groups used to filter graph data.
         """
 
-        # get column names of axes
-        x_column = x_axis.get("value")
-        y_column = y_axis.get("value")
-        # get units
-        # FIXME: add some check that all rows have the same units for each axis?
-        x_units = df[x_axis["units"]["column"]].iloc[0] if x_axis.get("units").get("column") \
-                  else x_axis.get("units").get("custom")
-        y_units = df[y_axis["units"]["column"]].iloc[0] if y_axis.get("units").get("column") \
-                  else y_axis.get("units").get("custom")
-        # determine axis labels
-        x_label = "{0}{1}".format(x_column.replace("_", " ").title(),
-                                  " ({0})".format(x_units) if x_units else "")
-        y_label = "{0}{1}".format(y_column.replace("_", " ").title(),
-                                  " ({0})".format(y_units) if y_units else "")
+        # get column names and labels for axes
+        x_column, x_label = get_axis_info(df, x_axis)
+        y_column, y_label = get_axis_info(df, y_axis)
 
         # find x-axis groups (series columns)
         groups = [x_column]
@@ -370,10 +359,9 @@ def read_perflog(path):
         df.drop(col, axis=1, inplace=True)
 
     # infer numeric columns
-    # null or "" becomes float, none becomes string
     df = df.apply(pd.to_numeric, errors="ignore")
     # set job completion time to datetime
-    df = df.astype({"job_completion_time": "datetime64[ns]"})
+    df["job_completion_time"] = pd.to_datetime(df["job_completion_time"])
 
     return df
 
@@ -405,6 +393,30 @@ def insert_key_cols(df: pd.DataFrame, index, results):
     for k in keys:
         # insert keys as new columns
         df.insert(index, k, [r[k] if k in r.keys() else "" for r in results])
+
+def get_axis_info(df: pd.DataFrame, axis):
+    """
+        Return the column name and label for a given axis. If a column name is supplied as units information, the actual units will be extracted from a dataframe.
+
+        df: dataframe, data to plot.
+        axis: dict, axis column and units.
+    """
+
+    # get column name of axis
+    col_name = axis.get("value")
+    # get units
+    units = axis.get("units").get("custom")
+    if axis.get("units").get("column"):
+        unit_set = set(df[axis["units"]["column"]])
+        # check all rows have the same units
+        if len(unit_set) > 1:
+            raise RuntimeError("Too many axis unit entries {0}".format(unit_set))
+        units = next(iter(unit_set))
+    # determine axis label
+    label = "{0}{1}".format(col_name.replace("_", " ").title(),
+                            " ({0})".format(units) if units else "")
+
+    return col_name, label
 
 def main():
 
--- a/post-processing/post_processing.py
+++ b/post-processing/post_processing.py
@@ -104,8 +104,35 @@ class PostProcessing:
         # apply user-specified types to all relevant columns
         for col in all_columns:
             if config["column_types"].get(col):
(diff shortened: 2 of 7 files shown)


post-processing/post_processing.py
```diff
@@ -164,6 +165,9 @@ def plot_generic(self, title, df, x_axis, y_axis, series_filters):
         for f in series_filters:
             if f[0] not in groups:
                 groups.append(f[0])
+        # all x-axis data treated as categorical
+        for g in groups:
+            df[g] = df[g].astype(str)
```
pineapple-cat, line 170: Might be a bit controversial if we want to reuse this filtered dataframe elsewhere for another graph type, but we could always get pandas to re-infer numeric types if necessary. As it stands, Bokeh needs its bar chart categories to be passed as strings.

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install .[test]
+          pip install .[test,post-processing]
```
pineapple-cat, line 40: Add optional dependency information to docs.

post-processing/post_processing.py
```diff
@@ -337,61 +339,45 @@ def read_perflog(path):
...
-
-    return pd.DataFrame.from_records(records)
+    # look for required column matches
+    required_field_matches = [len(list(filter(re.compile(rexpr).match, df.columns))) > 0 for rexpr in REQUIRED_LOG_FIELDS]
+    # check all required columns are present
+    if False in required_field_matches:
+        raise KeyError("Perflog missing one or more required fields", REQUIRED_LOG_FIELDS)
+
+    # replace display name
+    results = df["display_name"].apply(get_display_name_info)
```
pineapple-cat, line 358: Add unit test to ensure all display name params are collected.

tkoskela (changes requested):
Everything looks good to me with respect to #165 and #206

Could we add `Install` section  to `post-processing/README.md` that explains you install (with dependencies) by `pip install -e .[post-processing]` and that it requires python 3.9?

tkoskela (approved):

post-processing/post_processing.py
```diff
@@ -337,61 +339,45 @@ def read_perflog(path):
...
+    dict_cols = [c for c in ["extra_resources", "env_vars"] if c in df.columns]
+    for col in dict_cols:
+        results = df[col].apply(lambda x: ast.literal_eval(x))
+        # insert new columns and contents
+        insert_key_cols(df, df.columns.get_loc(col), results)
+        # drop old column
+        df.drop(col, axis=1, inplace=True)
+
+    # infer numeric columns
+    df = df.apply(pd.to_numeric, errors="ignore")
```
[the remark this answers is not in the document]
pineapple-cat: Date filtering works as expected, but different types of missing values result in different type inference behaviour from pandas. Columns that contain entirely `null`, `nan`, `NaN`, or `""` (empty field) are typed as float; everything else (`Null`, `NAN`, `none`, `None`) is typed as object (string). If you have a mixture of the two types of missing values (e.g. `null` and `Null`), automatic typing defaults to object.

It gets more interesting when you put regular values into the mix:
- A mixture of numeric and object-typed missing values will resolve as an object typed column.
- A mixture of string and float-typed missing values will resolve as an object typed column.
- A mixture of numeric and float-typed missing values will resolve as a float typed column.

post-processing/post_processing.py
ilectra: This is not a problem for now, but when the numeric/scaling plotting function is created in the future, you can have a line like this for the x-axis column, that converts it to numeric.

post-processing/post_processing.py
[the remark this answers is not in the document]
ilectra: This is getting too complicated. It would be more straightforward to add type in the config file for both the axis, all the filter columns, and all the series columns. And enforce that type here. the string->float will most likely need a custom function, in order to avoid type-errors.

post-processing/test_post_processing.py
```diff
@@ -237,6 +228,22 @@ def test_high_level_script(run_sombrero):
...
+        # FIXME: assert a certain column dtype for the extra param here?
+        assert e.args[1].columns.tolist() == EXPECTED_FIELDS
+    else:
+        assert False
+
+    # check correct date filtering
+    df = post_.run_post_processing(sombrero_changed_log_path, {"title": "Title", "filters": [["job_completion_time", ">", "2000-06-01T12:30:15"]], "series": [], "x_axis": {"value": "job_completion_time", "units": {"custom": None}}, "y_axis": {"value": "flops_value", "units": {"column": "flops_unit"}}})
+    # check returned subset is as expected
+    assert len(df) == 2
+
```
ilectra, line 246: Add a test that compares `datetime` with `date`, and is expected to fail with type error. In a new issue, we should fix this by converting any occurence of `date` in user-specified filters or series to `datetime`.

post-processing/post_processing.py
```diff
@@ -149,6 +150,7 @@ def plot_generic(self, title, df, x_axis, y_axis, series_filters):
         x_column = x_axis.get("value")
         y_column = y_axis.get("value")
         # get units
+        # FIXME: add some check that all rows have the same units for each axis?
```
ilectra, line 153: Yep! :)

post-processing/test_post_processing.py
pineapple-cat: I've discovered an additional complication; if the filter value of your date is interpreted as a string (i.e. you use something like `"2000-06-01"` in your config file), the filter comparison goes ahead and filters your data as expected. However, if the filter value of your date is `datetime.date` (i.e. you use `"2000-06-01"` in your config file), then you get a `TypeError`.

post-processing/post_processing.py
```diff
@@ -407,27 +379,45 @@ def get_display_name_info(display_name):
...
-    return columns
+    # get column name of axis
+    col_name = axis.get("value")
+    # get units
+    units = axis.get("units").get("custom")
+    if axis.get("units").get("column"):
+        unit_set = set(df[axis["units"]["column"]])
+        # check all rows have the same units
+        if len(unit_set) > 1:
+            raise RuntimeError("Too many axis unit entries {0}".format(unit_set))
```
pineapple-cat, line 414: When gathering the `unit_set`, the dataframe could likely use a `dropna()`; I noticed some logs that had rows with missing units. This won't work on anything pandas doesn't deem to be a missing value of course, and we'd have to change the `RuntimeError` check from `len(unit_set) > 1` to `len(unit_set) != 1`, but it seems reasonable to ignore missing units and type everything according to the rest of the rows.

post-processing/post_processing.py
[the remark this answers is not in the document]
pineapple-cat: Float columns allow for missing values, but int columns don't. For now I'll default to filling missing values in to-be-int columns with 0. Later we could perhaps let the user specify a fill value.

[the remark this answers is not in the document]
ilectra: Yes, use Int64 instead. But don't rely on the user to get the right type, change it internally.

post-processing/post_processing.py
ilectra: Yes!

post-processing/post_processing.py
```diff
@@ -104,8 +104,35 @@ def run_post_processing(self, log_path, config):
...
+                    conversion_type = "float64"
+                elif pd.api.types.is_integer_dtype(conversion_type):
+                    # all integers treated as Int64 (nullable)
+                    # note: default pandas integer type is int64 (not nullable)
+                    conversion_type = "Int64"
+                elif pd.api.types.is_datetime64_any_dtype(conversion_type):
+                    # all datetimes treated as datetime64[ns] (nullable)
+                    conversion_type = "datetime64[ns]"
+                else:
+                    raise RuntimeError("Unsupported user-specified type '{0}' for column '{1}'.".format(conversion_type, col))
```
pineapple-cat, line 128: Could also add a boolean type here. Don't think I've seen any perflogs that actually use bools, but pandas does have a nullable boolean type (missing values are treated as False).

ilectra: Don't worry about it for now, we can add `bool`s later if needed.

ilectra (approved):
Looks good, please merge!

PR #212, opened 2023-10-04
Allow less than max number of points per series

### LINKED ISSUE #211: Support different number of values per series
Currently plotting the following gives an error because we force all series to contain all x axis values:

```
x_axis:
  value: "test_name"
  units:
    custom: null

y_axis:
  value: "flops_value"
  units:
    column: "flops_unit"

filters: []

series: [["system", "archer2"], ["system", "isambard-macs"]]

```

```
RuntimeError: ('Unexpected number of rows (9) does not match number of unique x-axis values per series (12)',                  test_name  flops_value      flops_unit         system
0   HPCG_IntelSKX_Isambard      38.6863  Gflops/seconds  isambard-macs
1            HPCG_Original      24.0270  Gflops/seconds  isambard-macs
2             HPCG_Stencil      51.0070  Gflops/seconds  isambard-macs
3  HPCG_IntelAVX2_Isambard      39.0427  Gflops/seconds  isambard-macs
4               HPCG_LFRic      18.5399  Gflops/seconds  isambard-macs
5   HPCG_IntelAVX_Isambard      38.7652  Gflops/seconds  isambard-macs
6            HPCG_Original      39.2325  Gflops/seconds        archer2
7             HPCG_Stencil     124.1970  Gflops/seconds        archer2
8               HPCG_LFRic      55.9839  Gflops/seconds        archer2)
Post-processing stopped
```

We should be able to relax this to allowing the total number of values to be *less or equal* the number of series * the number of x axis values

tkoskela:
closes #211

### DIFF
--- a/post-processing/post_processing.py
+++ b/post-processing/post_processing.py
@@ -117,7 +117,7 @@ class PostProcessing:
         num_filtered_rows = len(df[mask])
         num_x_data_points = series_combinations * len(set(df[config["x_axis"]["value"]][mask]))
         # check expected number of rows
-        if num_filtered_rows != num_x_data_points:
+        if num_filtered_rows > num_x_data_points:
             raise RuntimeError("Unexpected number of rows ({0}) does not match number of unique x-axis values per series ({1})".format(num_filtered_rows, num_x_data_points), df[columns][mask])
 
         print("Selected dataframe:")


pineapple-cat (approved):
This can give us some slightly wonky-looking plots, but if we don't care about that the use case presented in your issue is valid and should be allowed.

PR #214, opened 2023-10-05
Keep `spack.lock` file after a successful benchmark finishes

### LINKED ISSUE #209: Request exclusive nodes on CSD3
Chris Edsall
  [1 hour ago](https://ukrse.slack.com/archives/C5JUM5A2E/p1695990174428659?thread_ts=1695987044.969689&cid=C5JUM5A2E)
Not on CSD3, we do tight packing of jobs on nodes. So for benchmarking it is necessary to request `--exclusive`

### LINKED ISSUE #210: Keep `spack.lock` file at the end of the job
This file can be useful to reproduce the spack build used for the benchmark run.

giordano:
I tested locally with the Sombrero benchmark and it seems to be doing the job, the file `spack.lock` is preserved in the output directory together with the other log files.

Note: in the benchmarks we should always append to `self.keep_files`, instead of forcibly overwriting its value. 

Fix #210.

### DIFF
--- a/benchmarks/apps/wrf/wrf.py
+++ b/benchmarks/apps/wrf/wrf.py
@@ -67,7 +67,7 @@ class WRFBaseBenchmark(SpackTest):
 
     executable = 'wrf.exe'
     executable_opts = []
-    keep_files = ['rsl.error.0000']
+    keep_files.append('rsl.error.0000')
 
     num_cpus_per_task = 2
     num_tasks = required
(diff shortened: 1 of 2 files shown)


tkoskela (approved):
Works for me!

I suppose the issue is that this will be overwritten along with other output if you rerun the same test. This is a bigger issue we should figure out at some point.

PR #220, opened 2023-10-09
Update Spack environment for Archer2

giordano:
Ref https://github.com/ukri-excalibur/excalibur-tests/pull/185#issuecomment-1634458867


giordano:
> Didn't go through all the updates, but the ones I looked at look good! I assume you had a look at https://github.com/EPCCed/epcc-spack/commit/543a2b68bf48e42d0ad25dbca6916c5c7b841129?

I didn't see that commit, thanks for pointing it out!  I compared the changes here to those there, we should be very similar.

PR #222, opened 2023-10-12
FEniCS Benchmark

chrisrichardson:
First draft of a benchmark for FEniCS, as sketched out at Excalibur Workshop in Bristol.


tkoskela:
For reference:

I tried running this on csd3. The concretized spack spec is
```
==> Concretized fenics-benchmark@main
 -   lnmdxvd  fenics-benchmark@main%gcc@11.2.0~ipo build_system=cmake build_type=Release generator=make arch=linux-centos7-skylake_avx512
 ```
the build fails with.
```
[ 18%] Generating Elasticity.c
ffcx /tmp/dc-kosk1/spack-stage/spack-stage-fenics-benchmark-main-lnmdxvdntyj2rrujjpkieu3ay3xxjzuf/spack-src/src/Poisson.py
make[2]: ffcx: No such file or directory
```
Does the spack package need to depend on `py-fenics-ffcx`?

**Comment by tkoskela** [comment]
Looks like we are missing python packages
```
    File "/home/dc-kosk1/demo/excalibur-tests/benchmarks/spack/csd3-cascadelake/compute-node/opt/linux-centos7-skylake_avx512/gcc-11.2.0/py-setuptools-63.4.3-fedflxl54t3m4ba4r2mudcwmitlhybua/lib/python3.8/site-packages/setuptools/_distutils/dist.py", line 992, in run_command
      cmd_obj.run()
    File "<string>", line 47, in run
    File "<string>", line 63, in build_extension
  ModuleNotFoundError: No module named 'nanobind'
  error: subprocess-exited-with-error
  
  × Building wheel for fenics-basix (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
```

chrisrichardson:
Looks like we are waiting for: https://github.com/spack/spack/pull/40440

tkoskela:
The PR has been merged, does this work now?

chrisrichardson:
I believe it should now work, yes. We probably need to tune some parameters before merging, though.

chrisrichardson:
Unfortunately, some other changes in spack have now caused it to fail again. I might lock to a version to prevent this keeping on happening.

giordano:
What's the problem specifically?

chrisrichardson:
A bunch of new dependencies for `py-fenics-basix@main` - it needs `py-scikit-build-core` now... 
I think more of this will keep happening, so probably best to stick to a release...

giordano:
Yeah, it's usually better to refer to fixed tags, rather than moving targets like branch names, which make reproducing benchmark runs at a different time much harder.

tkoskela:
Works on my local machine!

chrisrichardson:
I need to tidy up the actuall tests that are being run.

chrisrichardson:
Maybe we can merge this, as it seems to be working. I can work on making better scaling tests in another PR.

tkoskela:
Looks good to me, do you have any comments @giordano?

benchmarks/spack/repo/packages/fenics-benchmark/package.py
giordano, line 1: This is simple and small enough that it isn't a big deal to keep it here, but is there any chance this could be upstreamed to Spack?

giordano (approved):
Haven't tested it myself, but looks overall good to me, thanks!  Only a couple of minor comments.

Issue #205, opened 2023-09-05
Add OR and AND functionality to filtering


ilectra:
Mainly for row filters, but might be useful elsewhere as well.

Issue #183, opened 2023-07-11
Add data transformations in config and high-level script


ilectra:
In preparation for specific types of plots (especially scaling):
- allow `x_axis` or `y_axis` to be some transformation of the data, either between values of columns (eg. `tasks`/`num_CPUs`) or involving fixed values given in the config (eg. scaling by a fixed number). See if symPy can help with this.
- scale the y value for each series by a different number (eg. given in a dictionary). This is needed not only for scaling plots, eg. Babelstream benchmarks can be presented as scaled by the theoretical performance of each system.
- (largely moved to #194) scale y values by the value of the first point (usually the serial performance point in a scaling plot)

pineapple-cat:
- Could add ability to normalise ratios. (Is there a use-case for this?)
- Zeros and missing values throw exceptions.

PR #226, opened 2023-10-23
Add data transformations for post-processing plot data

pineapple-cat:
Addresses #183 and #205.

- [x] Fix categorical sorting.
- [x] Add unit tests.
- [x] Update documentation.

### DIFF
--- a/post-processing/README.md
+++ b/post-processing/README.md
@@ -41,6 +41,8 @@ Before running post-processing, create a config file including all necessary inf
     - `value` - Axis data points. Specified with a column name.
     - `units` - Axis units. Specified either with a column name or a custom label (may be null).
 - `filters` - (Optional.) Filter data rows based on specified conditions. (Specify an empty list if no filters are required.)
+  - `and` - Filter mask is determined from a logical AND of conditions in list.
+  - `or` - Filter mask is determined from a logical OR of conditions in list.
   - `Format: [column_name, operator, value]`
   - `Accepted operators: "==", "!=", "<", ">", "<=", ">="`
 - `series` - (Optional.) Display several plots in the same graph and group x-axis data by specified column values. (Specify an empty list if there is only one series.)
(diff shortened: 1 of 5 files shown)


pineapple-cat:
I still can't replicate exactly what's going wrong with Asif's example, but the scaling issue Tuomas found was caused by a `num_cores` mismatch, which I fixed by sorting the dataframe before scaling.

fixed_si_weak_scaling

asifsamiarain:
Thanks @pineapple-cat for the fix and now we may see the plot look like:

image

Still wonder, how to order w.r.t. series rather x-axis?

Here is below an example to give a quick look at the perflogs data:

```
import os
import glob
import pandas as pd
from pivottablejs import pivot_ui
 
path = os.getcwd()
files = glob.glob(os.path.join(path,"perflogs/*/*/*.log"))
df_list = (pd.read_csv(file, delimiter="|") for file in files)
df = pd.concat(df_list, ignore_index=True)
pivot_ui(df)
```

Above code will generate a browser viewable pivottablejs.html file (and above similar Sphng Single Node data w.r.t. 20230707, 20231013, 20231124 job completion times will look like grouped+ordered as shown below):

image

tkoskela (commented):
The documentation makes sense to me. A couple of suggestions:
- It might be easier to understand if the examples were in figures instead of tables.
- Instead of *A note on X* I would just have *X* as the subtitle.

While playing with Asif's logfiles, I noticed one more bit of odd behaviour. When I use the full datetime as the x axis, the series don't get sorted by `num_tasks_per_node` (or rather, it looks like they are ordered by the x axis value of the first element in the series).
image
If I drop the time from the datetimes, so that my series get grouped together on the x axis, I get the series sorted by `num_tasks_per_node`
image
Is this a bug or a feature?

Should we include some general formatting options in the config file? Something to think about for future. Things I always end up hacking by hand include
- orientation of the x axis labels
- export to png

pineapple-cat:
Bokeh has its own x-axis sorting method that I need to undermine at every step if we want non-string data to be sorted properly on a categorical plot. I've fixed this for (x, series) groupings in the commit below, but this will need to be revisited if we want to expand to (x, series1, series2) groupings. Here's what the plot should look like now:

#### Ascending
sphng_single_node_ascending
#### Descending
sphng_single_node_descending

Additionally, there's no need to hack anything to produce a PNG of the graph; this feature is available through the 'Save' button in the Bokeh toolbar:

bokeh_toolbar

ilectra (approved):

PR #227, opened 2023-10-27
Add CONQUEST benchmarks

tkoskela:
Add ReFrame class for Conquest, related to https://github.com/OrderN/CONQUEST-release/issues/264

### DIFF
--- a/benchmarks/apps/conquest/conquest_runonly.py
+++ b/benchmarks/apps/conquest/conquest_runonly.py
@@ -26,6 +26,14 @@ class ConquestBaseBenchmark(rfm.RunOnlyRegressionTest):
     def setup_variables(self):
         self.env_vars['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
 
+        if self.current_partition.scheduler.registered_name == 'sge':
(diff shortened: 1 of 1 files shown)


benchmarks/apps/conquest/conquest_runonly.py
```diff
@@ -0,0 +1,116 @@
...
+    input_dir = variable(str, value='foo')
+
+    # Build configuration
+    # spack_spec = 'conquest@develop +openmp'
+
+    @run_after('setup')
+    def setup_variables(self):
+        self.env_vars['OMP_NUM_THREADS'] = f'{self.num_cpus_per_task}'
+
+        if self.current_partition.scheduler.registered_name == 'sge':
```
giordano, line 29: Why do you need to do this here?

tkoskela: It's normally done in the `SpackTest` base class but this is a `Rfm.RunOnlyRegressionTest`

giordano: Ah, I missed you were deriving from `Rfm.RunOnlyRegressionTest`.

PR #233, opened 2023-11-08
Some improvements to CI setup

giordano:
* Use Dependabot to update CI workflows
* Set `RFM_CONFIG_FILES` once for the entire build
* Update `actions/checkout` to v4
* Clone Spack with `--depth=1`
* No need to reference `$Python_ROOT_DIR` for running the benchmarks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
tkoskela, line 7: What does this do, does it update the versions of the actions used in the workflow?

tkoskela (commented):
I didn't entirely understand what dependabot is doing, can you explain? Other changes look good.

tkoskela (approved):

Issue #237, opened 2023-11-24
Multiple issues with Sombrero benchmark


mirenradia:
When I do
```
reframe -c /path/to/excalibur-tests/benchmarks/apps/sombrero -l
```
I get the following warning (and no tests are found):
```
WARNING: skipping test file '/path/to/excalibur-tests/benchmarks/apps/sombrero/sombrero.py': module not found error: /path/to/excalibur-tests/benchmarks/apps/sombrero/sombrero.py:10: No module named 'apps'                                                                            
from apps.sombrero import case_filter 
```
Replacing `sombrero.py:L10` with
```python
from benchmarks.apps.sombrero import case_filter
```
resolves this issue but I then get the following warning
```
WARNING: skipping test file '/path/to/excalibur-tests/benchmarks/apps/sombrero/sombrero.py': reframe syntax error: variable 'valid_prog_environs' is declared in more than one of the parent classes of class 'SombreroBuild' (rerun with '-v' for more information)
```
I also have the same problem `SombreroBenchmarkBase`. I believe this is because `SombreroBuild` inherits from both `SpackTest` and `reframe.CompileOnlyRegressionTest` which both in turn inherit from `reframe.RegressionTest` (not sure how diamond inheritance works with Python).

I'm using the following versions
* Python 3.9.12
* ReFrame-HPC 4.4.1

Let me know if any further information would be helpful to debug this.

mirenradia:
Another issue is that for the `ITT-sn` and `ITT-64n` tests, `num_tasks` is set to use all of the cores on a node but this assumes the number of cores per node divides the lattice volume (32x24x24x24) which isn't true on some systems (e.g. the CSD3 Icelakes with 76 cores per node).

giordano:
* The problem `apps.sombrero` vs `benchmarks.apps.sombrero` might have been an overlook when we moved files into the new `benchmarks` subdirectory in #102
* I'll try to look into the inheritance issue
* Yes, these benchmarks should be more flexible and take into account the actual size of the node, or maybe it should at least allow to customise the volume on the command line, not sure there's an easy way to automagically guess a good lattice volume for an arbitrary number of cores

PR #239, opened 2023-11-27
[sombrero] Simplify classes used in benchmarks

giordano:
Address https://github.com/ukri-excalibur/excalibur-tests/issues/237#issue-2010042384.  @mirenradia Can you please check this solves the issues you reported in the first post?

### DIFF
--- a/benchmarks/apps/sombrero/sombrero.py
+++ b/benchmarks/apps/sombrero/sombrero.py
@@ -7,34 +7,18 @@ import reframe as rfm
 import reframe.utility.sanity as sn
 import reframe.utility.udeps as udeps
 
-from apps.sombrero import case_filter
+from benchmarks.apps.sombrero import case_filter
 from benchmarks.modules.reframe_extras import scaling_config
 from benchmarks.modules.utils import SpackTest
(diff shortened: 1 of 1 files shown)


mirenradia:
This resolves all of the issues I mentioned in #237 except for the `num_tasks` problem and I can now run the ITT-sn benchmark (after manually setting `num_tasks` appropriately). However, I do get the following warning:
```
WARNING: skipping test 'SombreroBenchmarkScaling': test has one or more undefined parameters
```

giordano:
I can reproduce that, as far as I understand it comes from the fact `case_filter.generate(scaling_config)` in https://github.com/ukri-excalibur/excalibur-tests/blob/40b4f4a874010ebfd822a46d08fcf1975020496d/benchmarks/apps/sombrero/sombrero.py#L103 is an empty list `[]`, and so the parameter `params` is undefined.  I'm not entirely sure how that benchmark is supposed to work, it was contributed by @mmesiti, maybe he can shed some light on this?

In any case I'm going ahead and merge this PR, as it addresses the previously reported issues.

PR #247, opened 2023-12-01
[CSD3] Update configurations and Spack environments

mirenradia:
This PR updates the ReFrame configurations for CSD3 and the corresponding Spack environments. Furthermore it also adds the new Sapphire Rapids partition.

Some key changes are explained below.

## Operating System

On CSD3, we have 2 different OSs:
* CentOS 7 - a rebuild of Red Hat Enterprise Linux (RHEL) 7
* Rocky Linux 8  - a rebuild of RHEL 8

The Cascade Lake partition run on CentOS 7 and all other partitions (i.e. Ice Lakes and Sapphire Rapids) run on Rocky Linux 8. In order to avoid bugs and library incompatibilities, it is important to use the software stack built for the correct OS. One of the main problems with the current `csd3-icelake` Spack environment is that there are lots of external packages built for CentOS 7/RHEL 7 and as a result I experienced segfaults on the example Sombrero benchmark with the default Spack spec. I have therefore completely recreated it from scratch and only included packages built for Rocky Linux 8.

In order to stress the different OSs to users of this repo on CSD3, I have renamed the existing systems as follows
* `csd3-cascadelake/compute-node` -> `csd3-centos7/cascadelake`
* `csd3-icelake/compute-node` -> `csd3-rocky8/icelake`

Note that there are two sets of login nodes `login-p-[0-4]` that are running Cent OS 7 and `login-q-[0-4]` that are running Rocky Linux 8.

## Sapphire Rapids Partition

This currently uses the same software stack and OS as the Ice Lakes hence the common `compilers.yaml` and `packages.yaml` for the Spack environment. The only difference in the Spack environments is the specified target.

The Sapphire Rapids partition has been added as a new partition (in the ReFrame sense) under the `csd3-rocky8` system.

## Preferred compilers and MPI implementation

I have set the preferred compiler to `intel` for all Spack environments and the preferred MPI implementation to `intel-mpi` (for CentOS 7) or `intel-oneapi-mpi` (for Rocky Linux 8). These are both in the default modules that are loaded on CSD3 and we generally observe better performance than with `gcc`/`openmpi`.

## Reference values

I have updated some of the existing reference values for the benchmarks.

I have tested this PR with the following benchmarks:
* grid
* hpl
* ramses
* sombrero
* sphng
* swift
* trove-pdsyev
* trove

There are still a few minor changes I wish to make before asking for a review hence why I'm making this a draft PR for now.


benchmarks/apps/hpl/csd3-rocky8/sapphirerapids/1/HPL.dat
giordano, line 1: Side note, I wish we could automate the generation of the input file, but at the moment we don't know the memory available on the compute nodes (it could be an extra custom property of partitions, if we wanted to), but there's also the problem that "looking up an ever-changing table on the Intel website) isn't exactly my idea of automation.

benchmarks/spack/csd3-centos7/cascadelake/spack.yaml
```diff
@@ -186,14 +332,22 @@ spack:
         prefix: /usr/local/software/hdf5/1.10.3
       - spec: hdf5@1.10.5%gcc@9.3.0+mpi^openmpi
         prefix: /usr/local/Cluster-Apps/hdf5/openmpi/gcc/9.2/1.10.5
-      - spec: hdf5@1.10.5%intel+mpi^intel-mpi
+      - spec: hdf5@1.10.5 +mpi^intel-mpi
```
giordano, line 335: I presume the point of `%intel` was not to allow using this with other libraries built with gcc, I may have had troubles trying this combination

benchmarks/reframe_config.py
```diff
@@ -87,32 +93,59 @@ def spack_root_to_path():
...
+                        ['I_MPI_OFI_PROVIDER', 'mlx'],
+                        ['UCX_NET_DEVICES', 'mlx5_0:1'],
+                    ],
                     'access': ['--partition=cclake', '--exclusive'],
                     'environs': ['default'],
                     'max_jobs': 64,
                     'processor': {
                         'num_cpus': 56,
-                        'num_cpus_per_core': 2,
+                        'num_cpus_per_core': 1,
```
giordano, line 142: Oops, good catch, thanks!

giordano (approved):
I haven't tested them myself, but the changes look good, thanks a lot!

benchmarks/apps/hpl/csd3-rocky8/sapphirerapids/1/HPL.dat
mirenradia: I agree that it would be nice to do this but don't think there's a good solution...

benchmarks/spack/csd3-centos7/cascadelake/spack.yaml
mirenradia: Yeah, for some reason Spack refused to used this external package (when using the `intel` compiler and `intel-mpi`) until I made this change. I didn't really understand it since it didn't seem to be a problem for any other external packages with compilers in their specs :shrug:

mirenradia:
I've rebased the changes onto `main`. This is now ready to be merged.

benchmarks/spack/csd3-centos7/cascadelake/spack.yaml
```diff
@@ -168,12 +244,82 @@ spack:
...
+      - spec: gmp@6.1.2
+        prefix: /usr/local/software/spack/spack-0.11.2/opt/spack/linux-rhel7-x86_64/gcc-5.4.0/gmp-6.1.2-sdvx5boug4isdzywdxgaqzwwor46yden
+    groff:
+      externals:
+      - spec: groff@1.22.2
+        prefix: /usr
+    libtool:
+      externals:
+      - spec: libtool@2.4.2
+        prefix: /usr
```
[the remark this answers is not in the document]
mirenradia: That makes sense. I've removed it.

mirenradia:
> Can you please sort alphabetically packages in the spack configurations? That makes easier to see what's in there.

It would be nice if `spack external find` did this automatically. In any case, I've done it now (by hand :see_no_evil:).

benchmarks/examples/sombrero/sombrero.py
```diff
@@ -101,7 +104,7 @@ class SombreroBenchmark(SpackTest):
         },
         '*': {
             'flops': (1, None, None, 'Gflops/seconds'),
-        },
+            },
```
giordano, line 107: I think this line was correctly indented before?
```suggestion
        },
```

giordano (approved):
Just a comment about indentation, but rest looks good, thanks again!

> It would be nice if spack external find did this automatically.

Agreed!

Issue #246, opened 2023-11-30
Sombrero ITT-sn benchmark is too small for the largest nodes


mirenradia:
Excluding the issues I identified in #237 and that have been partially fixed in #239. I've found that the Sombrero ITT-sn benchmark is too small for the largest nodes (in terms of number of cores). In particular, on the CSD3 Sapphire Rapids, with 112 cores per node (although I am only using 108 tasks due to [#237 (comment)](https://github.com/ukri-excalibur/excalibur-tests/issues/237#issuecomment-1827595717)), the results can fluctuate massively between repeated runs (even on the exact same node) - I have observed variations as much as ~80%.

Changing the arguments passed to the executable from `-s small` to `-s medium` resolve this problem for me but this is currently what is used for the 64 node benchmark. I don't have the resources to test this but I suspect this benchmark will also be too small for 64 nodes.

giordano:
CC: @mmesiti.

PR #250, opened 2023-12-04
[sombrero] Handle systems where number of cores doesn't divide lattice volume

giordano:
Should fix https://github.com/ukri-excalibur/excalibur-tests/issues/237#issuecomment-1827595717.  I followed the suggestion in https://github.com/ukri-excalibur/excalibur-tests/issues/237#issuecomment-1828323882.  Probably this isn't the most efficient algorithm out there, but this takes a bunch of microseconds on my laptop, so hopefully it won't be a major performance bottleneck:
```
In [7]: %%timeit
   ...: max_num_tasks(112, LATTICE_VOLUME)
   ...: 
   ...: 
3.75 µs ± 24.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

CC: @mirenradia.

### DIFF
--- a/benchmarks/apps/sombrero/sombrero.py
+++ b/benchmarks/apps/sombrero/sombrero.py
@@ -130,6 +130,7 @@ class SombreroITTsn(SombreroBenchmarkBase):
 class SombreroITT64n(SombreroBenchmarkBase):
     valid_systems = ['-gpu']
     tags = {"ITT-64n"}
+    num_nodes = variable(int, value=64, loggable=False)
 
     @run_after('init')
     def set_up_from_parameters(self):
@@ -139,6 +140,6 @@ class SombreroITT64n(SombreroBenchmarkBase):
     def setup_num_tasks(self):
         self.num_tasks_per_node = max_num_tasks(
             self.current_partition.processor.num_cores,
-            LATTICE_VOLUME_MEDIUM // 64,
+            LATTICE_VOLUME_MEDIUM // self.num_nodes,
         )
-        self.num_tasks = self.num_tasks_per_node * 64
+        self.num_tasks = self.num_tasks_per_node * self.num_nodes
--- a/benchmarks/apps/sombrero/sombrero.py
+++ b/benchmarks/apps/sombrero/sombrero.py
@@ -11,7 +11,8 @@ from benchmarks.modules.reframe_extras import scaling_config
 from benchmarks.modules.utils import SpackTest
 
 # Fixed lattice volume in ITT benchmarks
-LATTICE_VOLUME = 32 * 24 * 24 * 24
+LATTICE_VOLUME_SMALL = 32 * 24 ** 3
+LATTICE_VOLUME_MEDIUM = 48 ** 3 * 64
 
 
 # Helper function to find maximum number of tasks we can use for benchmarks
@@ -122,9 +123,8 @@ class SombreroITTsn(SombreroBenchmarkBase):
     def setup_num_tasks(self):
         self.num_tasks = max_num_tasks(
             self.current_partition.processor.num_cores,
-            LATTICE_VOLUME,
+            LATTICE_VOLUME_SMALL,
         )
-
 
 @rfm.simple_test
 class SombreroITT64n(SombreroBenchmarkBase):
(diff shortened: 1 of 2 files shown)


benchmarks/apps/sombrero/sombrero.py
```diff
@@ -122,4 +137,7 @@ def set_up_from_parameters(self):
 
     @run_after('setup')
     def setup_num_tasks(self):
-        self.num_tasks = self.current_partition.processor.num_cores * 64
+        self.num_tasks = max_num_tasks(
+            self.current_partition.processor.num_cores * 64,
```
giordano, line 141: I'm not sure whether `64` should multiply `self.current_partition.processor.num_cores` or the result of `max_num_tasks`.  In the latter case we'd greatly oversubscribe the node, right?  But that was already the case before, so probably that was it?

benchmarks/apps/sombrero/sombrero.py
```diff
@@ -5,12 +5,24 @@
...
 import reframe as rfm
 import reframe.utility.sanity as sn
-import reframe.utility.udeps as udeps
 
 from benchmarks.apps.sombrero import case_filter
 from benchmarks.modules.reframe_extras import scaling_config
 from benchmarks.modules.utils import SpackTest
 
+# Fixed lattice volume in ITT benchmarks
+LATTICE_VOLUME = 32 * 24 * 24 * 24
```
mirenradia, line 14: Looking at [the Sombrero README](https://github.com/sa2c/sombrero#problem-sizes), I think this is only the lattice volume in the case `-s small` is passed to the executable. For the 64 node test, `-s medium` is currently passed so the lattice volume is $48^3 \cdot 64$. It's possible these might be changed as well in order to resolve #246.

Having said that, I think this will probably still work fine since the prime factor decomposition of these numbers are both of the form $2^p3^q$ and they're much bigger than the number of cores on even the largest nodes.

benchmarks/apps/sombrero/sombrero.py
mirenradia: > In the latter case we'd greatly oversubscribe the node, right? But that was already the case before, so probably that was it?

I think this test was intended to run on 64 nodes so it shouldn't be oversubscribed. Maybe we should use the `num_tasks_per_node` option instead and calculate the `max_num_tasks` according to `LATTICE_VOLUME / 64` to ensure even load balancing and that we actually use 64 nodes?

mirenradia:
> Probably this isn't the most efficient algorithm out there, but this takes a bunch of microseconds on my laptop, so hopefully it won't be a major performance bottleneck.

I agree that this is extremely unlikely to be a problem so I have no issues with it.

benchmarks/apps/sombrero/sombrero.py
giordano: Thanks, I defined `LATTICE_VOLUME_SMALL` and `LATTICE_VOLUME_MEDIUM` to distinguish the two cases.

benchmarks/apps/sombrero/sombrero.py
```diff
@@ -122,4 +137,8 @@ def set_up_from_parameters(self):
 
     @run_after('setup')
     def setup_num_tasks(self):
-        self.num_tasks = self.current_partition.processor.num_cores * 64
+        self.num_tasks_per_node = max_num_tasks(
+            self.current_partition.processor.num_cores,
+            LATTICE_VOLUME_MEDIUM // 64,
+        )
+        self.num_tasks = self.num_tasks_per_node * 64
```
mirenradia, line 144: One very tiny comment. Could you replace the two uses of the magic number `64` with a variable e.g. `num_nodes`?

mirenradia (changes requested):
I'm happy with these changes other than my 1 small comment.

I have tested that the `ITT-sn` one works on the CSD3 Icelakes but unfortunately can't test the `ITT-64n` one as this exceeds the max job size (I did test it worked by replacing `64` with `48`).

benchmarks/apps/sombrero/sombrero.py
giordano: Done!  Can you please check it works for you as expected?

mirenradia:
LGTM (but I can't approve it until you re-request a review).

giordano:
Thanks!

PR #258, opened 2023-12-20
Refactor post-processing

### LINKED ISSUE #244: Refactor post-processing
Need to break down the master script into more modular pieces,that people can readily put together to make their own scripts.

pineapple-cat:
Addresses #244.

- [x] Perflog handler.
- [x] Config handler.
- [x] Modularise typing, sorting, filtering, and scaling.
- [x] Plot handler.

### DIFF
--- a/post-processing/plot_handler.py
+++ b/post-processing/plot_handler.py
@@ -122,7 +122,7 @@ def plot_generic(title, df: pd.DataFrame, x_axis, y_axis, series_filters, debug=
     plot.title.text_font_size = "15pt"
 
     # get label values with their original dtype
-    label_values = [pd.Series(x.label.value.split("=")[1].strip(), dtype=last_group_dtype).iloc[0]
+    label_values = [pd.Series(x.label["value"].split("=")[1].strip(), dtype=last_group_dtype).iloc[0]
                     for x in plot.legend[0].items]
     # sort legend items (order determined by x-axis sort)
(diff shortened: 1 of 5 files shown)


post-processing/plot_handler.py
```diff
@@ -0,0 +1,178 @@
...
+    # remove x-axis group ticks
+    plot.xaxis.major_tick_line_color = None
+    plot.xaxis.major_label_text_font_size = "0pt"
+    # FIXME: add this as a config option at some point
+    # plot.xaxis.group_label_orientation = "vertical"
+    # adjust font size
+    plot.title.text_font_size = "15pt"
+
+    # get label values with their original dtype
+    label_values = [pd.Series(x.label["value"].split("=")[1].strip(), dtype=last_group_dtype).iloc[0]
```
pineapple-cat, line 125: Specifically, I think changing `x.label.value` to `x.label["value"]` is needed to work with Bokeh v2.4.3, which is apparently the only Streamlit-compatible version. I haven't had the chance to test that this doesn't break anything with the current Bokeh v3.1.1, which is the version I was using until I put together the demo for #259.

ilectra: Good to keep a note of this! But don't change it here, do it in the PR that will introduce the need for it, which will be the one for streamlit.

ilectra (commented):
In pretty good shape! Please make sure all `FIXME`s are captured in issues.

ilectra (approved):

PR #260, opened 2024-01-23
Picking columns to export CSV

### LINKED ISSUE #252: Add more axes/dimensions/columns to post-processing
This doesn't need to include 3d plotting yet, but the PP needs the dataframe to be filtered based on 3 axis columns.

kaanolgu:
Fixes #252 

- For babelstream we need multiple columns to be exported so with this change,  the post-processing script would be able to recognise new column data alongside with x_axis and y_axis. If needed user could define a_axis, b_axis, etc. to export more columns and the default execution flow is not affected.
- Since the plotting part is not modified, it would not intervene the plotting part 
- For example, with this change we would be able to export a new column for dataframe : 
```python
x_axis:
  value: "tags"
  units:
    custom: "tags"
y_axis:
  value: "Triad_value"
  units:
    column: "Triad_unit"
z_axis:
  value: "spack_spec"
  units:
    column: null

series: [["partition", "cascadelake"],["partition", "volta"]]

```

- Then it would print ( commenting line `self.plot_generic( config["title"], df[columns][mask], config["x_axis"], config["y_axis"], series_filters)` [thanks @pineapple-cat ] ) : 
```
Selected dataframe:
   tags  Triad_value  Triad_unit                                    spack_spec    partition
0   acc    12963.337  MBytes/sec                   babelstream%gcc@13.1.0 +acc  cascadelake
1   acc    13386.104  MBytes/sec       babelstream%gcc@9.2.0 +acc cuda_arch=70        volta
2  cuda   846640.634  MBytes/sec      babelstream%gcc@9.2.0 +cuda cuda_arch=70        volta
3   omp   159131.926  MBytes/sec                   babelstream%gcc@13.1.0 +omp  cascadelake
4   tbb    99740.216  MBytes/sec  babelstream%gcc@13.1.0 +tbb partitioner=auto  cascadelake
```
 Is there any alternative method to export multiple columns that I might be missing ? Open to any feedback or suggestions 
 
 Thank you!


kaanolgu:
https://github.com/ukri-excalibur/excalibur-tests/commit/438be0257c39cf81bb1260e2a881559c2c10df3c

Made a change according to idea from @ilectra  and @pineapple-cat . Instead of adding a new "*_axis" to the `yaml` file, a new list of columns to extract to csv dataframe is used and `df_csv_export` is generated from the original dataframe. 

One question would be is it required to apply user-specified types to all relevant columns for the csv export too ?

pineapple-cat (commented):
I don't think we need user-specified types for csv export columns; typing is only needed for the parts of the dataframe that will be used in data manipulation in some way. Since what you want to add is purely for display purposes, it shouldn't make much difference once it's in the csv anyway.

ilectra:
I think that, instead of creating a whole new df to export to csv, it would make more sense if you
- add only the _extra_ columns that you need in the `csv_export` part of the yaml, not all the axis and series etc.
- keep those columns in the filtered dataframe, as you go along the processing/filtering, treating them as extra axes.
- export the filtered df to csv, just before calling the plotting script (you won't call it, in your case, but that's the place to do the printing)

kaanolgu:
> I think that, instead of creating a whole new df to export to csv, it would make more sense if you
> 
> * add only the _extra_ columns that you need in the `csv_export` part of the yaml, not all the axis and series etc.
> * keep those columns in the filtered dataframe, as you go along the processing/filtering, treating them as extra axes.
> * export the filtered df to csv, just before calling the plotting script (you won't call it, in your case, but that's the place to do the printing)

That's a great idea! I will try to implement this in a new commit

ilectra:
> > I think that, instead of creating a whole new df to export to csv, it would make more sense if you
> > 
> > * add only the _extra_ columns that you need in the `csv_export` part of the yaml, not all the axis and series etc.
> > * keep those columns in the filtered dataframe, as you go along the processing/filtering, treating them as extra axes.
> > * export the filtered df to csv, just before calling the plotting script (you won't call it, in your case, but that's the place to do the printing)
> 
> That's a great idea! I will try to implement this in a new commit

@kaanolgu you might want to wait for the refactoring PR to be merged. It will change things quite a bit, hopefully to the better!

post-processing/post_processing.py
```diff
@@ -23,12 +23,13 @@ def __init__(self, log_path: Path, debug=False, verbose=False):
         # dataframe filters
         self.mask = pd.Series(self.df.index.notnull())
 
-    def run_post_processing(self, config: ConfigHandler):
+    def run_post_processing(self, log_path: Path, config: ConfigHandler):
```
pineapple-cat, line 26: I have one more note about this - it would be preferable to no longer need a `log_path` as an input to the `run_post_processing` function since refactoring specifically moved it to only be used by `init`. Since this is now needed here again, perhaps we can just store it in the class during `init` and use `self.log_path` in `run_post_processing`?

kaanolgu: Actually, it's not a must to export the `.csv` file under the `perflogs` folder but I thought it would be good to have everything related with perflogs in the same folder. 

The easiest solution is to export it to the directory we run the benchmark I guess without any modification to functions?

pineapple-cat: Yes, that also works, but as you said it might be better to keep perflog-related things together in one folder. It's completely fine to use `self.log_path` in `run_post_processing` (just throw a `self.log_path = log_path` into `init` since we need a `log_path` to initialise the `PostProcessing` class anyway and we know the path has to exist), I only wanted to remove it as an argument because it makes stitching together the front-end more convenient.

post-processing/post_processing.py
```diff
@@ -50,6 +51,10 @@ def run_post_processing(self, config: ConfigHandler):
         # FIXME (#issue #255): have an option to put this into a file (-s / --save flag?)
         print("Selected dataframe:")
         print(self.df[self.mask][config.plot_columns])
+        if self.debug:
+            print("CSV dataframe:")
+            print(self.df[self.mask][config.plot_columns + config.extra_columns])
+        self.df[self.mask][config.plot_columns + config.extra_columns].to_csv(str(self.log_path)+'/output.csv', index=True)  # Set index=False to exclude the DataFrame index from the CSV
```
pineapple-cat, line 57: I did a quick investigation into why the tests are failing: you're trying to concatenate `self.log_path` with `"output.csv"`, but that won't always work because the `log_path` isn't necessarily a directory, it can also be a file. I think the best way to fix this is that we can either try to find an appropriate parent folder in the path, or let the user specify their own `output_path` where they'd like to save the CSV.

post-processing/post_processing.py
ilectra: I think we should keep all the output in one place, and not mix it with the perflogs. This place is currently the directory where we run the benchmark. If we want to control the output location, we can add an `output_path` to the postprocessing arguments - but please open a separate issue for this!

post-processing/post_processing.py
ilectra: As I said above, I prefer the former option :)

post-processing/post_processing.py
```diff
@@ -50,6 +51,10 @@ def run_post_processing(self, config: ConfigHandler):
         # FIXME (#issue #255): have an option to put this into a file (-s / --save flag?)
         print("Selected dataframe:")
         print(self.df[self.mask][config.plot_columns])
+        if self.debug:
+            print("CSV dataframe:")
+            print(self.df[self.mask][config.plot_columns + config.extra_columns])
```
ilectra, line 56: Please don't duplicate lines 52-53, just add the `+ config.extra_columns` to the `print` statement in line 53- making sure it doesn't error if `config.extra_columns` doesn't exist...

post-processing/post_processing.py
ilectra, line 57: Do we want to control the printing of the csv with a flag, or print it every time?

post-processing/post_processing.py
```diff
@@ -62,64 +67,15 @@ def run_post_processing(self, config: ConfigHandler):
...
-        # FIXME: add scaling for x-axis
-        scaling_columns = []
-        # extract scaling columns
-        if config["y_axis"].get("scaling"):
-            if config["y_axis"]["scaling"].get("column"):
-                scaling_columns.append(config["y_axis"]["scaling"]["column"]["name"])
-
-        # gather all relevant columns
-        all_columns = set(columns + filter_columns + scaling_columns)
+        return self.df[config.plot_columns][self.mask]
```
ilectra, line 70: I assume this mess here is leftover from the merging with the refactored version?

post-processing/post_processing_config.yaml
```diff
@@ -49,8 +49,6 @@ column_types:
   cpus_per_task: "int"
 
 # Specify which columns to export to csv file
-csv_export:
-  [tasks,
-  flops_value,
-  Triad_unit,
+additional_columns_to_csv:
+  [flops_value,
```
ilectra, line 53: This is not an additional column, it's the `y_axis`, therefore it will be printed out anyway. I think you're mixing two different configs. 
This config file is just an example, so you can either add an imaginary column name here, that is not mentioned anywhere else in the config, or make it vaguely correct for the Sombrero perflog, currently sitting in the `tutorial` folder, and pick a random column from there - again, it shouldn't be mentioned anywhere else in the config.

ilectra (commented):
Please also add unit tests for the new functionality!

post-processing/post_processing.py
ilectra: Opened issue #306

ilectra:
@pineapple-cat @kaanolgu I think I addressed all my review comments, please have a last look and merge if happy!

pineapple-cat (approved):
Everything seems to be in order to me :thumbsup:

Issue #251, opened 2023-12-07
Parser for the spack_spec


kaanolgu:
Performance Portability post-processing will need the spack_spec parsed, to fish out the compiler name (the thing after `%`) and the variants (the thing after `+` is an on-off, which for babelstream is called language, and there are custom ones as well with `variant=value`).

ilectra:
It'll be safer to do this with the spack api, and not as a string parsing. But accessing the spack API is a PITA, so @giordano has some ideas of how to do it. 
Ideally, the parsing of the spack_spec will be done during the benchmarking run, and then its components will be saved in separate columns in the perflog.

giordano:
I have already a proof of concept of doing something like this in #35.  The main ingredients are:

* in `benchmarks/modules/utils.py`, append to the attribute `self.postrun_cmds` (note: _append_, instead of overwriting the attribute as done in that PR, that wasn't a good idea) of the `SpackTest` class a command which prints out whatever you want
* in the same file, add to the `SpackTest` class a method decorated with `@run_after('run')` which captures the quantity you printed out with the post run command above and save into a new attribute of the class
* add the value of the variable captured in this way to the perflog

### How to print out the desired information

In #35 I was running the command `$(spack -e <path to environment> build-env <spack spec> | grep -oP '^SPACK_CC=\K.*) --version`, but that was a very clunky way to do it.  A better way is to use [Spack Python API](https://spack.readthedocs.io/en/latest/spack.html) (see for example [spack tutorial about scripting](https://spack-tutorial.readthedocs.io/en/latest/tutorial_spack_scripting.html)).  Spack isn't usually installed as a regular Python package, so accessing its module isn't straightforward short of fiddling with `PYTHONPATH`, which can be very fragile, but Spack provides the command `spack python` to run a Python interpreter which knows how to access the `spack` module.  For example, to show the list of specs in the active environment, assuming it has been already concretized (which is the case for the environment we use in our case) you can do something like
```
spack -e <path to environment> python -c 'from spack import environment; [x for x in environment.active_environment().added_specs()]'
```
In practice, you'll need to tweak the command to print exactly what you want.

### Caveats

There are a few caveats to be aware of:

* the string you print out should be unique enough that the extract value in postrun command with `sn.extractsingle` doesn't accidentally captures something else
[...]

PR #262, opened 2024-02-02
Parsing compiler name & version from Spack environment to include in the performance log file

kaanolgu:
Attempt for issue https://github.com/ukri-excalibur/excalibur-tests/issues/251
Thanks: @giordano


benchmarks/apps/babelstream/babelstream.py
giordano, line 1: As outlined in https://github.com/ukri-excalibur/excalibur-tests/issues/251#issuecomment-1914696265, all that you're doing here should be done in `benchmarks/modules/utils.py` for the `SpackTest` class.

kaanolgu: Addressed this in https://github.com/ukri-excalibur/excalibur-tests/pull/262/commits/323ae1142337563783d5de5ca04ac5ae2ad9d883

benchmarks/reframe_config.py
```diff
@@ -752,6 +752,8 @@ def spack_root_to_path():
                         '%(check_environ)s|'
                         '%(check_extra_resources)s|'
                         '%(check_env_vars)s|'
+                        '%(check_compiler_name)s|'
+                        '%(check_compiler_version)s|'
```
ilectra, line 756: Don't you also want the variant fished out of the `spack_spec`?

kaanolgu: Right now, seems like not needed anymore since `tags` from ReFrame has the programming model variant, and compiler name alongside with version is provided with this PR. In the new Spack package we have made some changes to make sure each programming model has its own dedicated flags. So, it is tough to have a one solution fits all for the rest of the variants for Babelstream and also for other apps too since naming conventions are all different.

benchmarks/modules/utils.py
```diff
@@ -267,10 +269,22 @@ def setup_spack_environment(self):
             f'(cd {cp_dir}; find . \( -name "spack.yaml" -o -name "compilers.yaml" -o -name "packages.yaml" \) -print0 | xargs -0 tar cf - | tar -C {dest} -xvf -)',
             f'spack -e {self.build_system.environment} config add "config:install_tree:root:{env_dir}/opt"',
         ]
-
+        cmd_compiler_name = 'from spack import environment; print(environment.active_environment().spec_lists["specs"].specs[0].compiler.name)'
+        cmd_compiler_version = 'from spack import environment; environment.active_environment().spec_lists["specs"].specs[0].compiler.versions[0]'
```
ilectra, line 273: There can be more than one package in the spec, so instead of the `[0]` element, we should write out a dictionary for each package. Something like `... | {'compiler_name_package1':'name1', 'compiler_name_package2':'name2', ...} | 'compiler_version_package1':'version1', 'compiler_version_package2':'version2', ...} | <other parts of the spec> ...| The postprocessing would then have to be minimally modified to be able to parse this, and create a separate column for each key of each dictionary.

benchmarks/reframe_config.py
ilectra: Then maybe we should export the whole variant? And users of each app can choose to use it and unpack it or not.

giordano:
Besides compilers, another specific bit of information that users may be interested in is the MPI implementation used, which can be fished out with `spec["mpi"]`, with the caveat that not all packages depend on MPI, so this would need to check whether `"mpi" in spec`, so not to error out (e.g. `spec["mpi"] if "mpi" in spec else ""`).

kaanolgu:
for ` spack install babelstream%gcc@13.1.0 +std build_system=cmake std_submodel=data` :

```
$ spack -e /lustre/home/br-kolgu/excalibur-tests-upstream/stage/isambard-macs/cascadelake/default/STDDATABenchmark/spack_env/cascadelake python -c 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants;print(d)'
+std build_system=cmake std_submodel=data
```
According to the [manual](https://spack.readthedocs.io/en/latest/spack.html#spack.variant.VariantMap) `dict` should do the job:
```
$ spack -e /lustre/home/br-kolgu/excalibur-tests-upstream/stage/isambard-macs/cascadelake/default/STDDATABenchmark/spack_env/cascadelake python -c 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants;print(d.dict)'
{'std': BoolValuedVariant('std', True), 'build_system': AbstractVariant('build_system', 'cmake'), 'std_submodel': AbstractVariant('std_submodel', 'data')}
```
Keys : 
```
$ spack -e /lustre/home/br-kolgu/excalibur-tests-upstream/stage/isambard-macs/cascadelake/default/STDDATABenchmark/spack_env/cascadelake python -c 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants;print(d.dict.keys())'
dict_keys(['std', 'build_system', 'std_submodel'])
```

From here we try : 
```
$ spack -e /lustre/home/br-kolgu/excalibur-tests-upstream/stage/isambard-macs/cascadelake/default/STDDATABenchmark/spack_env/cascadelake python -c 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants;print(d.dict["std_submodel"])'
std_submodel=data

$ spack -e /lustre/home/br-kolgu/excalibur-tests-upstream/stage/isambard-macs/cascadelake/default/STDDATABenchmark/spack_env/cascadelake python -c 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants;print(d.dict["std_submodel"].value)'
('data',)

$ spack -e /lustre/home/br-kolgu/excalibur-tests-upstream/stage/isambard-macs/cascadelake/default/STDDATABenchmark/spack_env/cascadelake python -c 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants;print(d.dict["std_submodel"].value[0])'
data
```
So we could form a loop to run through all the keys and export them as different columns for the perflog. But my question was during the meeting for the `reframe_config.py` file :
https://github.com/ukri-excalibur/excalibur-tests/blob/323ae1142337563783d5de5ca04ac5ae2ad9d883/benchmarks/reframe_config.py#L733-L758

how do we know the names ? I have an idea to export all the spack_spec variants into a list and run through that list in the `reframe_config.py` file to print each column is it valid approach ?

kaanolgu:
This modified code for the `utils.py` : 
```
        spack_spec_keys = 'from spack import environment; list(environment.active_environment().spec_lists["specs"].specs[0].variants.dict.keys())'
        spack_spec_vals = 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants.dict; l = list(d.keys()); values=[d[key].value[0] if isinstance(d[key].value,tuple) else d[key].value for key in l];print(values) '
        cmd_compiler_name = 'from spack import environment; print(environment.active_environment().spec_lists["specs"].specs[0].compiler.name)'
        cmd_compiler_version = 'from spack import environment; environment.active_environment().spec_lists["specs"].specs[0].compiler.versions[0]'
        self.postrun_cmds.append(f'echo "compiler_name: $(spack -e {self.build_system.environment} python -c \'{cmd_compiler_name}\')"')
        self.postrun_cmds.append(f'echo "compiler_version: $(spack -e {self.build_system.environment} python -c \'{cmd_compiler_version}\')"')
        self.postrun_cmds.append(f'echo "Spack_Spec keys : $(spack -e {self.build_system.environment} python -c \'{spack_spec_keys}\')"')
        self.postrun_cmds.append(f'echo "Spack_Spec vals : $(spack -e {self.build_system.environment} python -c \'{spack_spec_vals}\')"')
```
Prints out : 
```
compiler_name: gcc
compiler_version: 13.1.0
Spack_Spec keys : ['std', 'build_system', 'std_submodel']
Spack_Spec vals : [True, 'cmake', 'ranges']
```
For  : `spack install babelstream%gcc@13.1.0 +std build_system=cmake std_submodel=data`

I will check if sanity library function [extractall](https://reframe-hpc.readthedocs.io/en/v2.8/sanity_functions_reference.html#reframe.utility.sanity.extractall) is what we need for reading all the values from keys to pass to the `reframe_config.py` and values to be assigned for each. 

I am open for any other suggestions if there is anything more direct approach than this

ilectra:
I think you're almost there, @kaanolgu . The output I was thinking (for one package in the spec, see my comment above about many packages) is something like:
```
Spack_Spec : {'std':True, 'build_system':'cmake', 'std_submodel':'ranges'}
```
instead of the two separate lists `spack_spec-keys` and `spack_spec-vals`. I think a `dict(zip(spack_spec-keys,spack_spec-vals))` should do it.

kaanolgu:
@giordano @ilectra 
https://github.com/ukri-excalibur/excalibur-tests/pull/262/commits/8ba75611d79834c3a0d4d6083cce21c90e9345dc Tested with following :
Input:
```
reframe -c benchmarks/apps/babelstream -r --tag cuda --system=isambard-macs:volta --setvar=num_cpus_per_task=40 -S build_locally=false -Sspack_spec='babelstream%gcc@9.2.0,12.1.0 +cuda cuda_arch=70,72'
```
Output :
```
spack_spec: {'cuda': True, 'cuda_arch': {'70', '72'}, 'compiler_name': 'gcc', 'compiler_version': {'12.1.0', '9.2.0'}}
```

Output (perflog) : 
```
...|spack_spec|...
...|{'cuda': True, 'cuda_arch': {'70', '72'}, 'compiler_name': 'gcc', 'compiler_version': {'9.2.0', '12.1.0'}}|...
```

kaanolgu:
> Besides compilers, another specific bit of information that users may be interested in is the MPI implementation used, which can be fished out with `spec["mpi"]`, with the caveat that not all packages depend on MPI, so this would need to check whether `"mpi" in spec`, so not to error out (e.g. `spec["mpi"] if "mpi" in spec else ""`).

Just saw this, will add it in a new commit

pineapple-cat (approved):
The added columns don't seem to require any additional changes to post-processing, so this looks good to me on that front.

benchmarks/modules/utils.py
```diff
@@ -267,11 +271,38 @@ def setup_spack_environment(self):
...
+        self.postrun_cmds.append(f'echo "MPI: $(spack -e {self.build_system.environment} python -c \'{cmd_spec_mpi}\')"')
+        
         # Keep the `spack.lock` file in the output directory so that the Spack
         # environment can be faithfully reproduced later.
         self.keep_files.append(os.path.realpath(os.path.join(self.build_system.environment, 'spack.lock')))
 
+    @run_after('run')
+    def get_compiler_name(self):
+        with osext.change_dir(self.stagedir):
+            self.compiler_name = sn.extractsingle(r'compiler_name: \s*(\S+)', self.stdout, 1).evaluate()
```
ilectra, line 292: Do we need to do this as a two step process? (write out the information to stdout before compilation and then read it from stdout after running) Is there a way to do it in one go, at the right time for ReFrame?

benchmarks/modules/utils.py
```diff
@@ -267,11 +271,38 @@ def setup_spack_environment(self):
             f'(cd {cp_dir}; find . \( -name "spack.yaml" -o -name "compilers.yaml" -o -name "packages.yaml" \) -print0 | xargs -0 tar cf - | tar -C {dest} -xvf -)',
             f'spack -e {self.build_system.environment} config add "config:install_tree:root:{env_dir}/opt"',
         ]
-
+        # else f"{d[key].value}" if isinstance(d[key].value, bool) for boolean with "True" instead of True
+        cmd_spack_variants = 'from spack import environment; d = environment.active_environment().spec_lists["specs"].specs[0].variants.dict; keys = list(d.keys());values = {key: ([d[key].value[i] for i in range(len(d[key].value))] if isinstance(d[key].value, tuple) else d[key].value) for key in keys};print(values)'
```
ilectra, line 275: Is this the whole variant string broken down into keys automatically or as one string?

ilectra: Actually.... If this contains the compiler name and version as well, do we still need the separate variables and commands to fish them out? Can we let the post-processing unpack it into separate columns instead?

ilectra, line 275: In all these commands, you extract only the first package from the spec with `print(environment.active_environment().spec_lists["specs"].specs[0]` . There should be a dictionary _for each package_ instead.

ilectra:
> The added columns don't seem to require any additional changes for post-processing to function normally, so this looks good to me on that front.
> 
> Edit: Do note, however, that if you'd like to unpack any of your dict fields, please see how `extra_resources` and `env_vars` are handled in lines 119-125 of the `read_perflog` function in `perflog_handler.py`.

Yes, you have to unpack those the way Emily says. And once you add all the packages in the spec, it will be a bit more complicated than `extra_resources` and `env_vars`, as your values will not be just a dict anymore, but a nested dict of dicts (see my review comments and other comments above).
Also, in order to make the postprocessing not dependent on perflog version, there should be a check if those columns exists, before it's attempted to unpack them.

benchmarks/modules/utils.py
kaanolgu: I thought it would be better to keep variants and compiler information separate since compiler name and version is _always_ available but some applications won't even have a variant at all maybe and spack_spec as a string is available in the columns as a whole anyway.

ilectra: I think it's not multiple specs, but multiple packages inside the one spec, see eg. https://github.com/ukri-excalibur/excalibur-tests/blob/0a60516fcc73c8914984bc7e36fbd739bb0caea2/benchmarks/apps/hpl/hpl.py#L17 
Do I get this right @giordano?

giordano: > Do I get this right @giordano?

Almost, they are two distinct top-level specs in one environment.  But the point is that, as I had already explained in the second point of the caveats section of https://github.com/ukri-excalibur/excalibur-tests/issues/251#issuecomment-1914696265, you can't assume there's exactly 1 top-level spec in the environment, which is what you're doing when you do `specs[0]`.

giordano (changes requested):

benchmarks/modules/utils.py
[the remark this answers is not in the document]
kaanolgu: I think I found it from [documentation](https://spack.readthedocs.io/en/latest/spack.environment.html#spack.environment.Environment.concretized_user_specs): 
```
cmd_spack_disp = 'from spack import environment; d = environment.active_environment().concretized_user_specs;print(d);'
self.postrun_cmds.append(f'echo "DISPLAY: $(spack -e {self.build_system.environment} python -c \'{cmd_spack_disp}\')"')

prints out : 

DISPLAY: [babelstream%gcc@9.2.0+cuda cuda_arch=70,72]
```
I could run an iteration ( instead of just picking 0) and create dictionaries for each _concretized_ spec ?
 @ilectra I think this would also address the issue you mentioned in your comment [here](https://github.com/ukri-excalibur/excalibur-tests/pull/262#discussion_r1512745801) ?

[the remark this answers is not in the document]
ilectra: Yeah, the loop you're describing sounds right. I've no idea of the thing Mose is talking about.

Remember to adjust the postprocessing accordingly when you add the loop!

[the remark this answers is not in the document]
kaanolgu: https://github.com/ukri-excalibur/excalibur-tests/pull/262/commits/f696cd55a0fb2205e4e1923c197061bb134dc09d
@giordano @ilectra 
Input : ( `intel-tbb` and `cuda` are unrelated but I couldn't get `hpl` working on Isambard so I just wanted to test multiple packages in one spec) 
```
$ reframe -c benchmarks/apps/babelstream -r --tag cuda --system=isambard-macs:volta --setvar=num_cpus_per_task=40 -S build_locally=false -Sspack_spec='babelstream%gcc@9.2.0 +cuda cuda_arch=70,72 intel-tbb%gcc@9.2.0'
```

Output : 
```
spack_spec_dict: {'babelstream': {'compiler': {'name': 'gcc', 'version': [9.2.0]}, 'variants': {'cuda': True, 'cuda_arch': ('70', '72')}}, 'intel-tbb': {'compiler': {'name': 'gcc', 'version': [9.2.0]}, 'variants': {}}, 'mpi': ()}
```
so the dictionary is now more easy to access since we would have ` ["babelstream"]["compiler"]["name"] `
Also, if mpi is not present in all list of specs, it will be an empty tuple. I kept `mpi` out in the first level dictionary since it is something extra so anyone who needs to acess `mpi` data just finds it via ["mpi"][i] where i will be the package order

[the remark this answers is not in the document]
kaanolgu: #TODO : (for me ) 
- babelstream_compiler_name 
- babelstream_compiler_version
- babelstream_variant
- babelstream_mpi 

4 columns

From perflog_handler.py:
https://github.com/ukri-excalibur/excalibur-tests/blob/c2caf65ae61a08b719b2300e90e62e271593ce46/post-processing/perflog_handler.py#L118C1-L125C43

Add spack_spec_dict to here and then use another loop to parse one level down ( do not parse variants since we don't know for each app )

benchmarks/modules/utils.py
```diff
@@ -267,11 +268,25 @@ def setup_spack_environment(self):
             f'(cd {cp_dir}; find . \( -name "spack.yaml" -o -name "compilers.yaml" -o -name "packages.yaml" \) -print0 | xargs -0 tar cf - | tar -C {dest} -xvf -)',
             f'spack -e {self.build_system.environment} config add "config:install_tree:root:{env_dir}/opt"',
         ]
-
+        cmd_spack_spec_dict =    'from spack import environment;\
+                                spec_list = environment.active_environment().concretized_user_specs;\
```
giordano, line 272: We debugged this together offline, I think this should be [`environment.active_environment().concrete_roots()`](https://spack.readthedocs.io/en/latest/spack.environment.html#spack.environment.Environment.concrete_roots):
```suggestion
                                spec_list = environment.active_environment().concrete_roots();\
```

benchmarks/modules/utils.py
```diff
@@ -267,11 +268,25 @@ def setup_spack_environment(self):
...
+                                key_list_for_each = [[*spec_list[i].variants.dict.keys()] for i in range(len(spec_list))];\
```
giordano, line 273: I think this can be simplified to
```suggestion
                                key_list_for_each = [spec.variants.dict.keys() for spec in spec_list];\
```

kaanolgu:
@giordano @ilectra 
https://github.com/ukri-excalibur/excalibur-tests/pull/262/commits/49f22867d29feafd1886ac7dea08bf3e9e79244b
Fixes the issues we discussed and also includes the post-processing too

tkoskela:
The pytests pass on my local machine, but for some reason the GHA ci workflow still fails. We are wondering if it's using a cached version of `utils.py` that is not writing the spec into the output.

Issue #270, opened 2024-02-20
Add config upload and download to Streamlit UI


pineapple-cat:

ilectra:
Closed via #273

PR #273, opened 2024-02-23
Add config upload and download to Streamlit UI for post-processing

pineapple-cat:
Addresses #270.

### DIFF
--- a/post-processing/streamlit_post_processing.py
+++ b/post-processing/streamlit_post_processing.py
@@ -71,7 +71,13 @@ def update_ui(post: PostProcessing, config: ConfigHandler):
     st.multiselect("Series", config.series if config.series else [None], config.series,
                    placeholder="None", label_visibility="collapsed", disabled=True)
 
-    st.button("Generate Graph", on_click=rerun_post_processing)
+    generate_graph, download_config = st.columns(2)
+    with generate_graph:
+        st.button("Generate Graph", on_click=rerun_post_processing, use_container_width=True)
+    with download_config:
+        st.download_button("Download Config", config.to_yaml(),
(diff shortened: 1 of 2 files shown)


post-processing/streamlit_post_processing.py
```diff
@@ -71,7 +73,13 @@ def update_ui(post: PostProcessing, config: ConfigHandler):
     st.multiselect("Series", config.series if config.series else [None], config.series,
                    placeholder="None", label_visibility="collapsed", disabled=True)
 
-    st.button("Generate Graph", on_click=rerun_post_processing)
+    generate_graph, download_config = st.columns(2)
+    with generate_graph:
+        st.button("Generate Graph", on_click=rerun_post_processing, use_container_width=True)
+    with download_config:
+        st.download_button("Download Config", config.to_yaml(),
```
ilectra, line 80: Let the user choose where to save the file when pressing this button.

ilectra: Also, the style of the generated yaml is not exactly the same as our input one. This has no effect on functionality, but it might be jarring to the user to see that the yaml file they downloaded doesn't look the same as the one they're used to writing for the standalone plotting, for example. If it doesn't take too much time, please investigate how to bring those two to be more similar!

ilectra (approved):

post-processing/streamlit_post_processing.py
pineapple-cat: #275 now opened.

pineapple-cat: After a bit of investigation, it seems like there's no out-of-the-box solution for selecting where your downloaded files go. We could create an issue for this, but it doesn't seem useful to focus on adding this here.

PR #277, opened 2024-03-01
Add configuration to limit parallel local jobs to 1

ilectra:
This should solve the problem of spack having to lock files while building and then the lock timing out and failing the build.

I think we saw this problem during the CIUK23 tutorial on ARCHER2.


giordano (approved):

ilectra:
Yes, we've done it for the GitHub Actions runner, but not for any of the clusters.

Issue #275, opened 2024-02-26
Consistency between config yaml examples and downloaded configs from Streamlit


pineapple-cat:
"The style of the generated yaml is not exactly the same as our input one. This has no effect on functionality, but it might be jarring to the user to see that the yaml file they downloaded doesn't look the same as the one they're used to writing for the standalone plotting, for example. If it doesn't take too much time, please investigate how to bring those two to be more similar!"

Issue #269, opened 2024-02-20
Activate filters and series in Streamlit UI


pineapple-cat:
**Bonus**:
  - [ ] ~~Create custom tag-like component for filter/series display and removal.~~

PR #278, opened 2024-03-04
Activate filters and series in Streamlit UI for post-processing

pineapple-cat:
Addresses #269.

- [x] Filter/series addition.
- [x] Filter/series removal.
- [x] Redundant column type removal.

**Bonus:**
- [x] YAML dump format improvement for config download. (#275)

### DIFF
--- a/post-processing/config_handler.py
+++ b/post-processing/config_handler.py
@@ -119,7 +119,7 @@ class ConfigHandler:
         """
             Convert information in the class to a yaml format.
         """
-        return yaml.dump(self.to_dict(), sort_keys=False)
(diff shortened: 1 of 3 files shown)


ilectra:
Looks pretty good! Just a couple of comments:
- [ ] Can the df not be written out into the console every time the plot is updated? If it's too hard to do this, don't worry about it.
- [ ] The `update_ui` function is getting a bit big. See if you can modularise it a bit/move bits of it into separate functions, and add some comments in the code of what each section does.

ilectra (approved):

PR #280, opened 2024-03-07
Add kathleen config

ilectra:
The sombrero example runs fine. The conquest one doesn't , so probably the spack environment needs tweaking, but the conquest benchmark is itself under development, so I don't want to delay merging this system because of it. Discuss.


tkoskela:
Yeah I wouldn't worry too much if a benchmark in a draft PR (#227) doesn't run properly. However, it seemed like there was an issue with `openmpi`?

ilectra:
Can you do a last review and merge if happy with it, @tkoskela and @giordano ? Both intel-mpi and OpenMPI work with the current versions in the `spack.yaml`.

benchmarks/reframe_config.py
```diff
@@ -157,6 +157,39 @@ def spack_root_to_path():
...
+                    'descr': 'Kathleen compute nodes',
+                    'scheduler': 'sge',
+                    'launcher': 'mpirun',
+                    'environs': ['default'],
+                    'max_jobs': 36,
+                    'processor': {
+                        'num_cpus': 80,
+                        'num_cpus_per_core': 2,
+                        'num_sockets': 2,
+                        'num_cpus_per_socket': 20,
```
tkoskela, line 178: I feel like this should represent the physical, not virtual cores. If a benchmark wants to run on a "full node" it would grab the `num_cpus` number and request that many cores per node. I assume with `-l threads=2` it will get rejected by the scheduler? I guess I need to go and test this. But since we are disabling hyperthreading by default I think the processor config should reflect this.

benchmarks/reframe_config.py
```diff
@@ -157,6 +157,39 @@ def spack_root_to_path():
...
+                    },
+                    'resources': [
+                        {
+                            'name': 'mpi',
+                            'options': ['-pe mpi {num_slots}'],
+                        },
+                        {
+                            'name': 'threads',
+                            'options': ['-l threads=2'], # disable hyperthreading (default). To enable, use threads=1
```
tkoskela, line 187: Can you do this on the command line? Editing this file to toggle this is not ideal.

benchmarks/spack/kathleen/compute-node/spack.yaml
```diff
@@ -0,0 +1,320 @@
...
+      - spec: libxsmm@1.16.1
+        prefix: /shared/ucl/apps/cp2k/8.2/cp2k-8.2/tools/toolchain/install/libxsmm-1.16.1
+    m4:
+      externals:
+      - spec: m4@1.4.16
+        prefix: /usr
+    ninja:
+      externals:
+      - spec: ninja@1.10.2.git.kitware.jobserver-1
+        prefix: /shared/ucl/apps/python/bundles/python39-6.0.0/venv
```
tkoskela, line 273: Your python was complaining about this yesterday when you changed the version to 3.11, wasn't it? Does this package implicitly depend on python 3.9, as the name might suggest.

tkoskela (commented):
I made some comments but I think I actually need to do my homework and test these things.

benchmarks/spack/kathleen/compute-node/spack.yaml
ilectra: No, it wasn't. It might if one uses the package, but I haven't...

benchmarks/reframe_config.py
giordano: > But since we are disabling hyperthreading by default 

What do you mean exactly?

benchmarks/reframe_config.py
ilectra: ok, I misunderstood how the resource should be declared/used. How's this:

```suggestion
                            'options': ['-l threads={virtual_cpus_per_thread}'], # default is 2 to disable hyperthreading. To enable, use {'threads' : {'virtual_cpus_per_thread': 1}}
```

ilectra: The above is the value for the `self.extra_resources` inside a benchmark that would want to enable hyperthreading.

giordano: Maybe this should be controlled via [`use_multithreading`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.use_multithreading)?  At the moment the SGE scheduler doesn't do anything special in ReFrame about that, but if desired we could do something like https://github.com/reframe-hpc/reframe/blob/2b289c55e1c1015dd7f2dd04fdeac33f3a30305d/reframe/core/schedulers/slurm.py#L199-L202 for Slurm (as long as SGE has some generic enough configurations for this).

benchmarks/reframe_config.py
tkoskela: > Are you suggesting we should set `num_cpus_per_core=1` and then `num_cpus=40`?

Yes, I think we should. We do the same e.g. on Archer2 where we set 

https://github.com/ukri-excalibur/excalibur-tests/blob/3a9f51fe0b275c5ba75b4d1ece92ceef61e4fb7c/benchmarks/reframe_config.py#L60-L64

even though the number of CPU(s) reported by `lscpu` is 256 with 2 Threads per core.

At least we should be consistent on how we set this across our systems.

benchmarks/reframe_config.py
tkoskela: I agree it would be great if this could be handled by ReFrame and controlled by `use_multithreading` in a benchmark.

PR #282, opened 2024-03-13
Create resource group to apply memory setting for Isambard Instinct

giordano:
@kaanolgu can you please check this works for you?  If I remember correctly, we need to define the `extra_resource` so that it's applied in the configuration, but in this case we don't need (yet?) to define any custom variable, so I think an empty dictionary should work?

### DIFF
--- a/benchmarks/modules/utils.py
+++ b/benchmarks/modules/utils.py
@@ -294,9 +294,9 @@ class SpackTest(rfm.RegressionTest):
     def set_isambard_instinct_memory(self):
         # We want to define the `instinct_memory` extra resource, so that we
         # can use the corresponding resources in the config file.
-        system, partition = current_partition.fullname.split(':')
-        if system == "isambard-phase3" and partition == "instinct":
-            self.extra_resources['instinct_memory'] = {}
+        system, partition = self.current_partition.fullname.split(':')
+        if system == "isambard-phase3":
+            self.extra_resources['memory'] = {}
 
     @run_after('setup')
     def setup_build_job_num_cpus(self):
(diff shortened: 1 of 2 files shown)


benchmarks/modules/utils.py
```diff
@@ -290,6 +290,14 @@ def set_sge_num_slots(self):
             # Set the total number of CPUs to be requested for the SGE scheduler.
             self.extra_resources['mpi'] = {'num_slots': num_tasks * num_cpus_per_task}
 
+    @run_before('compile')
+    def set_isambard_instinct_memory(self):
+        # We want to define the `instinct_memory` extra resource, so that we
+        # can use the corresponding resources in the config file.
+        system, partition = current_partition.fullname.split(':')
```
kaanolgu, line 297:
```suggestion
        system, partition = self.current_partition.fullname.split(':')
```

benchmarks/modules/utils.py
```diff
@@ -290,6 +290,14 @@ def set_sge_num_slots(self):
...
+        if system == "isambard-phase3" and partition == "instinct":
```
kaanolgu, line 298:
```suggestion
        if system == "isambard-phase3":
```

giordano:
The reason why we haven't dealt with the memory as a proper resource so far is because it's complicated: it's a mixture of application-specific (memory requirements can vary _a lot_) and system-specific (I think for some schedulers it's the memory per CPU, in others it's the total memory), so it's really hard to come up with reasonable default values which work well in all situations.

giordano:
@kaanolgu is it good to go now?

kaanolgu:
> @kaanolgu is it good to go now?

LGTM

PR #293, opened 2024-03-21
[docs] Exclude unrelated files

giordano:
`apps/` and `post-processing/` are symlinks to the directories in our repository, with the result the website includes *all* the files in them.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -71,16 +71,15 @@ jobs:
         run: |
           if [[ ${{ github.event_name }} == 'pull_request' ]]; then
             export MKDOCS_SITE_DIR="site/preview/PR${{ github.event.number }}"
-          else
-            export MKDOCS_SITE_DIR="site"
+            export MKDOCS_SITE_URL="https://ukri-excalibur.github.io/excalibur-tests/preview/PR${{ github.event.number }}"
           fi
           mkdocs build
 
       - name: Deploy docs
         # Run only if push is to `main`, or if it's a PR not from a fork.
         if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') || (github.event_name == 'pull_request' && ! github.event.pull_request.head.repo.fork) }}
+        working-directory: gh-pages
         run: |
-          cd gh-pages
           cp -fr ../site/* .
           git config user.name ${{github.actor}}
           git config user.email "${{github.actor_id}}+${{github.actor}}@users.noreply.github.com"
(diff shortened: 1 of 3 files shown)


giordano:
Our website currently bundles _all_ files in `apps/` and `post-processing`, with the result the [entire website has a size of whopping 39.2 MiB](https://github.com/ukri-excalibur/excalibur-tests/actions/runs/8377880301).  Instead with this PR the entire website, including the preview for the current PR, is [39.9 MiB](https://github.com/ukri-excalibur/excalibur-tests/actions/runs/8379106337), which means the website in the preview is only 0.7 MiB, instead of 39.2 MiB.

giordano:
An alternative is to selectively symlink the `README.md` files for each individual benchmark, but that'd add an extra step to adding a new benchmark.

tkoskela:
The `Details` link on the Documentation deployed action seems to have broken

giordano:
Right, because that's a Python script, not a bash one 😅  Will fix it.

giordano:
Fixed, now the link should be correct again.  For reference, it's https://ukri-excalibur.github.io/excalibur-tests/preview/PR293/

.github/workflows/docs.yml
```diff
@@ -61,29 +61,46 @@ jobs:
...
+          # cleanly update the website.
           if [[ ${{ github.event_name }} == 'pull_request' ]]; then
-              COMMIT_MESSAGE="Update docs preview from ${{ github.sha }} for PR #${{ github.event.number }}"
+              COMMIT_MESSAGE="Update docs preview from ${{ github.sha }}"
+              # Only delete files in the preview subdir, if any
+              git rm -rf "${PREVIEW_SUBDIR}"
           else
               COMMIT_MESSAGE="Update docs from ${{ github.sha }}"
+              for file in $(git ls-files); do
+                  if [[ x"${file}" != xpreview/* ]]; then
```
tkoskela, line 96: I don't understand this bit. You want to delete all files in the repository, except ones starting with `preview/`? Is the `x` just enforcing that `preview` is at the start of the path?

giordano: Yes, and yes. The leading "x" is a common shell trick to ensure the comparison is done correctly, autotools scripts use it extensively for example.

tkoskela (approved):
Thanks for spotting this before we bloated the git history too badly!

giordano:
We might also want to squash the history of the gh-pages branch at some point, so that we forget about the files that had been checked in.

giordano:
New website is only [691 KiB](https://github.com/ukri-excalibur/excalibur-tests/actions/runs/8389577101), and [this diff](https://github.com/ukri-excalibur/excalibur-tests/commit/bd1f87764dfc8e34e8afdaccd25b53a071f8a618) shows all the deleted files 🙂

PR #297, opened 2024-03-25
Update spack version in CI and add note in docs

### LINKED ISSUE #296: Update spack requirement in docs and CI
#262 fails on spack 0.18.0 in the [CI](https://github.com/ukri-excalibur/excalibur-tests/actions/runs/8420122591) but passes with the latest develop. On my local machine tests pass with spack 0.21.0. The documentation doesn't mention a minimum spack version anymore (I suppose it used to?). 

We should update the version in CI and keep the same version in the documentation.

tkoskela:
Closes #296

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,8 @@ jobs:
         os:
           - ubuntu-latest
         spack_version:
-          # In README.md we claim compatibility with v0.21.0
-          - v0.21.0
+          # In README.md we claim compatibility with v0.20.0
+          - v0.20.0
           - develop
         spack_spec:
           # `example` is a package in our custom Spack repo, ensure we can build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,8 @@ jobs:
         os:
           - ubuntu-latest
         spack_version:
-          # In README.md we claim compatibility with v0.18.0
-          - v0.18.0
+          # In README.md we claim compatibility with v0.21.0
+          - v0.21.0
           - develop
         spack_spec:
           # `example` is a package in our custom Spack repo, ensure we can build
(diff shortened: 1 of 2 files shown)


kaanolgu:
@tkoskela  I just checked it is the `conrete_roots` function that we call, it seems it was introduced in the version `v.20.0` so checking against `v.21.0` is fine but could also be `v.20.0` if we want to check backwards compatibility 
https://spack.readthedocs.io/en/latest/spack.environment.html#spack.environment.Environment.concrete_roots [OK]
https://spack.readthedocs.io/en/v0.20.0/spack.environment.html#spack.environment.Environment.concrete_roots [OK]
https://spack.readthedocs.io/en/v0.19.2/spack.environment.html#spack.environment.Environment.concrete_roots [NOK]

docs/install.md
```diff
@@ -81,6 +81,11 @@ Check spack is in `$PATH`, for example
 spack --version
 ```
 
+### Version
+
+We recommend using spack v0.21.0 or later. Older versions of spack may work but are not
```
ilectra, line 86:
```suggestion
We recommend downloading locally and using the latest spack version (develop). If a stable version is needed, that has to be spack v0.21.0 or later at the time of writing. Older versions of spack may work but are not 
```

tkoskela: I think the downloading part is covered earlier. Agree about recommending the latest. How about

https://github.com/ukri-excalibur/excalibur-tests/blob/170aef375bfd6f4d157d9e42c09d32f012164abf/docs/install.md?plain=1#L86-L87

.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           - ubuntu-latest
         spack_version:
           # In README.md we claim compatibility with v0.21.0
```
ilectra, line 46: Oops, the comment still says v0.21.0...

kaanolgu (approved):

PR #305, opened 2024-04-08
[CI] Do not rewrite links in symlinked files

### LINKED ISSUE #304: Broken URLs in preview docs
The URL rewriting in #295 does not seem to work as intended. I get in https://ukri-excalibur.github.io/excalibur-tests/preview/PR299/ url's on the main page that look like `https://ukri-excalibur.github.io/excalibur-tests/preview/PR299/preview/PR299/install/` ie. the `preview/PR299` bit appears twice.

giordano:
If we rewrite the hyperlinks in symlinked files we risk to accidentally append duplicate `${PREVIEW_SUBDIR}`, depending on the order with which `find` operates on the files.

Should fix #304.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -75,7 +75,8 @@ jobs:
           echo "BASE_URL=${BASE_URL}" >> "${GITHUB_ENV}"
           PREVIEW_SUBDIR="preview/PR${{ github.event.number }}"
           echo "PREVIEW_SUBDIR=${PREVIEW_SUBDIR}" >> "${GITHUB_ENV}"
-          find . -name '*.md' -print -exec sed -i "s|${BASE_URL}|${BASE_URL}/${PREVIEW_SUBDIR}|g" '{}' \;
+          # Edit only files, and not symlinks, to avoid double editing the same files.
+          find . -type f -name '*.md' -print -exec sed -i "s|${BASE_URL}|${BASE_URL}/${PREVIEW_SUBDIR}|g" '{}' \;
 
       - name: Build docs
         run: |


giordano:
@tkoskela I'm not crazy, right?  [This line](https://github.com/ukri-excalibur/excalibur-tests/actions/runs/8601157311/job/23567812326#step:7:285) after rewriting the links is
```md
- [Installation](https://ukri-excalibur.github.io/excalibur-tests/preview/PR305/install/)
```
right?  But then the generated website has the duplicate subdir
https://github.com/ukri-excalibur/excalibur-tests/blob/5d49476b4fee1aa9a437c9eb72037a0c689c0ec2/preview/PR305/index.html#L1212

giordano:
As far as I understand: the link rewriting command _is_ working as intended, but mkdocs _sometimes_ is doing something stupid.  I say sometimes because in #295 [it was working correctly](https://github.com/ukri-excalibur/excalibur-tests/issues/304#issuecomment-2042770836).

giordano:
After a long time staring at the screen to spot the error, I believe now I understand the problem: the `find` command is editing all `*.md` files, but some of them are symlinks, and so there's the risk to rewrite the links twice if the original file is modified first, and the symlink is modified afterwards (once the symlink is modified by `sed` it's turned into a real file and the symlink is broken, if the order is inverted we wouldn't get double rewrite of the same file).

This probably also explains why the error was occurring occasionally: the occurrence of the error depends on the order with which `find` operates on the files, which is probably unspecified and can change from run to run.

tkoskela (approved):

PR #318, opened 2024-05-02
Delete `benchmarks/apps/legacy` directory

giordano:
If we ever need stuff from this directory we can also bring it back from the history.


tkoskela (approved):
:boom:

PR #325, opened 2024-05-23
Add line plots

connoraird:
Adding the option to plot line charts.


post-processing/streamlit_post_processing.py
```diff
@@ -68,6 +68,14 @@ def update_ui(post: PostProcessing, config: ConfigHandler, e: 'Exception | None'
         # config file uploader
         st.file_uploader("Upload Config", type="yaml", key="uploaded_config", on_change=update_config)
 
+        # set plot type
+        plot_type_options = ['generic', 'line']
+        plot_type_index = plot_type_options.index(config.plot_type) if config.plot_type else None
```
pineapple-cat, line 73:
```suggestion
        plot_type_index = plot_type_options.index(config.plot_type) if config.plot_type else 0
```
As plot type is required now, I'd make the fallback index 0 so that in the absence of information we default to a generic plot.

post-processing/plot_handler.py
```diff
@@ -10,6 +10,77 @@
...
+
+    # adjust axis ranges
+    min_y = (0 if min(df[y_column]) >= 0
+             else math.floor(np.nanmin(df[y_column])*1.2))
+    max_y = (0 if max(df[y_column]) <= 0
+             else math.ceil(np.nanmax(df[y_column])*1.2))
+    min_x = (0 if min(df[x_column]) >= 0
+             else math.floor(np.nanmin(df[x_column])*1.2))
+    max_x = (0 if max(df[x_column]) <= 0
+             else math.ceil(np.nanmax(df[x_column])*1.2))
```
pineapple-cat, line 38: Note: This will not work for datetime x-axis values. Not sure if that's out of scope for this PR, but might be worth leaving a `#FIXME` comment somewhere.

post-processing/plot_handler.py
```diff
@@ -10,6 +10,77 @@
...
+    """
+    # get column names and labels for axes
+    x_column, x_label = get_axis_labels(df, x_axis, series_filters)
+    y_column, y_label = get_axis_labels(df, y_axis, series_filters)
+
+    # adjust axis ranges
+    min_y = (0 if min(df[y_column]) >= 0
+             else math.floor(np.nanmin(df[y_column])*1.2))
+    max_y = (0 if max(df[y_column]) <= 0
+             else math.ceil(np.nanmax(df[y_column])*1.2))
```
pineapple-cat, line 34: These automatic y-axis endpoints look fine for bar charts, but many line charts will contain mostly empty space with a small line far above 0. I don't know how often users will want the y-axis to start at 0 as opposed to just giving them a clearer closeup view of the data they want to see, but perhaps it might be helpful to make this more dynamic?

pineapple-cat (commented):
Looking good so far! Don't forget to update any unit test that has a config with `"plot_type": "generic",` at the start so we can make sure the CI is passing cleanly.

post-processing/streamlit_post_processing.py
connoraird: I went for `None` so that the placeholder text is displayed and the user would then choose the plot type but I'm happy to default to generic if that's preferred?

pineapple-cat: I think it is; the point of generic is that it should be able to display any other type of chart, so even if the user forgets to switch to a different plot type they'll get to see an output instead of a red screen.

post-processing/plot_handler.py
```diff
@@ -12,6 +13,29 @@
...
+        datetime_range = axis_max_element - axis_min_element
+        buffer_time = datetime_range*0.2
+        axis_min = axis_min_element - buffer_time
+        axis_max = axis_max_element + buffer_time
+    else:
+        if not (axis_min and axis_max):
+            axis_min = (axis_min_element*0.6 if min(df[column]) >= 0
+                    else math.floor(axis_min_element*1.2))
+            axis_max = (axis_max_element*0.6 if max(df[column]) <= 0
+                    else math.ceil(axis_max_element*1.2))
```
pineapple-cat, line 35: Good idea, but having tried this in practice custom ranges don't work if you use 0 as one of your values (which is a valid value to want for either your min or max), so I think you might need a more specific check for this if statement.

We also get some interesting behaviour if you set your min value to be greater than your max value: the plot gets mirrored, which isn't necessarily a bad thing as long as it's not breaking anything, but definitely worth making a note of this somewhere. A tooltip maybe? Not sure if Streamlit allows you to attach tooltips to checkboxes though...

post-processing/streamlit_post_processing.py
```diff
@@ -177,6 +177,16 @@ def axis_select(label: str, axis: dict):
...
     units_select(label, axis)
+    
+    # axis range
+    use_default_ranges = st.checkbox("Use default ranges", axis.get("range").get("use_default"), key="{0}_axis_range_use_default".format(label))
+    if not use_default_ranges:
+        axis_range_min, axis_range_max = st.columns(2)
+        with axis_range_min:
+            st.number_input("Minimum", key="{0}_axis_range_min".format(label))
+        with axis_range_max:
+            st.number_input("Maximum", key="{0}_axis_range_max".format(label))
```
pineapple-cat, line 188: From a UX standpoint it's a little unclear to have these min/max options available for generic graphs when they do precisely nothing for any plot that isn't a line graph at the moment. It's not a major problem and doesn't need to be addressed in this PR, but worth either opening a new issue or adding to an existing "fix UI" issue when this PR gets merged.

connoraird: Good point, I believe you can dynamically hide them depending if plot_type = line is selected. Would that work?

pineapple-cat: Yes, that's what I was thinking.

post-processing/plot_handler.py
connoraird: Ah that is annoying about zero not being allowed. but you can have (+-)0.01 which could possibly cover the needed functionality, although i agree it's not ideal and a tooltip would definitely be needed.

I suppose the mirroring axes is just the same as the sort descending functionality ,so it might not be so bad?

connoraird: I tried to do this but couldn't get it working do I'll add a FIXME comment

pineapple-cat:
This has remained undocumented thus far, but I've been using `flake8` with a 120 char/line limit for style enforcement. Perhaps it's time to add a linter to the CI...

post-processing/streamlit_post_processing.py
```diff
@@ -169,6 +177,23 @@ def axis_select(label: str, axis: dict):
 
     # units select
     units_select(label, axis)
+
+    # FIXME: add ability to use a custom value for only one of min or max
+    disable_user_range = plot_type != 'line'
```
pineapple-cat, line 182:
```suggestion
    disable_user_range = plot_type != "line" or axis.get("value") is None
```
I noticed an error today where the web app isn't happy if you try to set a custom range before selecting which axis column you'd like to use. I reckon the easiest way to deal with this is to just not allow the scenario to happen and keep custom ranges disabled until an axis column is selected.

post-processing/streamlit_post_processing.py
```diff
@@ -169,6 +177,23 @@ def axis_select(label: str, axis: dict):
...
+                                     disabled=disable_user_range,
+                                     help="{0} {1}".format(
+                                         "Uncheck to assign custom minimum and maximum values to axis.",
+                                         "Custom ranges are not implemented for datetime types."))
+    if not use_default_ranges:
+        axis_range_min, axis_range_max = st.columns(2)
+        with axis_range_min:
+            st.number_input("{0}-axis minimum".format(label), key="{0}_axis_range_min".format(label))
+        with axis_range_max:
+            st.number_input("{0}-axis maximum".format(label), key="{0}_axis_range_max".format(label))
```
pineapple-cat, line 195:
```suggestion
        with axis_range_min:
            st.number_input("{0}-axis minimum".format(label),
                            value=(axis["range"].get("min") if axis.get("range").get("min")
                                   else get_axis_min_max(df[mask], axis)[0]),
                            key="{0}_axis_range_min".format(label))
        with axis_range_max:
            st.number_input("{0}-axis maximum".format(label),
                            value=(axis["range"].get("max") if axis.get("range").get("max")
                                   else get_axis_min_max(df[mask], axis)[1]),
                            key="{0}_axis_range_max".format(label))
```
Yesterday I thought it would be nice if the initial values of these number inputs were to correspond to the default axis values generated, but I didn't quite have enough time to craft a proper addition. I think something like what I've put in the suggestion above would work, but I've only had the chance to test it haphazardly and I know that using `axis.get(...).get(...)` can result in unwanted behaviour.

I'd also recommend adding in a line for `mask = st.session_state.post.mask` somewhere at the top of this function, preferably under `df = st.session_state.post.df`, as well as a `from plot_handler import get_axis_min_max` above the `PostProcessing` import.

If this turns out to be too much of a pain to get working it can become a FIXME instead.

post-processing/streamlit_post_processing.py
```diff
@@ -169,6 +177,23 @@ def axis_select(label: str, axis: dict):
 
     # units select
     units_select(label, axis)
+
+    # FIXME: add ability to use a custom value for only one of min or max
+    disable_user_range = plot_type != 'line'
+    use_default_ranges = st.checkbox("default axis range",
+                                     axis.get("range").get("use_default") if not disable_user_range else True,
```
pineapple-cat, line 184: Minor nitpick, but is this `use_default` bool for the axis ranges particularly useful anywhere else? It's technically information we can derive (i.e. `use_default = axis.get("range").get("min") is None and axis.get("range").get("max") is None`) and might feel like a slightly redundant field for anyone creating a plot config the manual way.

pineapple-cat (commented):
Good work, very close to being done I think. I've just made a few more notes and noticed that a good final touch to add would be to update the README config structure and config template sections to account for the new custom axis ranges.

pineapple-cat (approved):
I think I'm more or less happy with the way this PR looks now. The important functionality is all there and I don't see any more major bugs that need to be fixed. Pytest is still passing locally.

Once this gets merged, I think any of the following bits that aren't implemented here should collectively belong to some sort of line plot enhancement issue:

- Update README config structure + template sections to account for new custom axis ranges.
- Implement custom axis ranges for bar charts OR hide range selectors. (For now could also add tooltips to range selectors with this information.)
- Allow use of custom values for only one of min or max.

post-processing/config_handler.py
```diff
@@ -280,8 +297,12 @@ def read_config(config: dict):
 
     # check optional series information
     if config.get("series"):
-        if len(config.get("series")) == 1:
-            raise RuntimeError("Number of series must be >= 2.")
+        if plot_type == "generic":
+            if len(config.get("series")) == 1:
+                raise RuntimeError("Number of series must be >= 2 for generic plot.")
+        if plot_type == "line":
```
ilectra, line 303: Why would this be different for line plots?
It looks like it makes more sense to have `len(series)>=1` instead of 2, but that should be investigated in a different issue. Here just keep the same limit for both types of plots (if nothing breaks)

ilectra (approved):
Need to open an issue to disable log scale for x axis for `generic` type plots.

post-processing/config_handler.py
ilectra: The test for series probably needs changing as well.

Issue #327, opened 2024-06-07
Add new benchmark : Arepo


gokmenkilic:
- [ x ] Write a Spack Recipe

This recipe only include **yee_2d** simulation, we need to add variants for other simulations.

I have tried to add this but it did not worked for me:

```
variant(
        'config', 
        default='yee_2d', 
        values=(
            'wave_1d', 
            'shocktube_1d',
            'polytrope_1d_spherical',
            'noh_3d',
            'noh_2d',
            'mhd_shocktube_1d',
            'isolated_galaxy_collisionless_3d',
            'interacting_blastwaves_1d',
            'gresho_2d',
            'galaxy_merger_star_formation_3d',
            'current_sheet_2d',
            'cosmo_zoom_gravity_only_3d',
            'cosmo_box_star_formation_3d',
            'cosmo_box_gravity_only_3d',
            'alfven_wave_1d'
        ),
        description='Choose config file'
    )
```

- [x] Write a Reframe Test

PR #328, opened 2024-06-07
Add new benchmark: Arepo 

gokmenkilic:
This PR address the enhancements in issue #327 .


benchmarks/apps/arepo/arepo.py
```diff
@@ -0,0 +1,47 @@
...
+import reframe as rfm
+import reframe.utility.sanity as sn
+from benchmarks.modules.utils import SpackTest
+import os.path as path
+
+@rfm.simple_test
+class ArepoBenchmark(SpackTest):
+    valid_systems = ['-gpu']
+    valid_prog_environs = ['default']
+    spack_spec = 'arepo@master'
```
tkoskela, line 15: Building from the head of a branch is not recommended because it breaks reproducibility. There is no way to know later which commit of master was built. This is mentioned in [the documentation](https://ukri-excalibur.github.io/excalibur-tests/contributing/#spack-package). I'd normally recommend building off a tag or a release, but the repository does not seem to have any :facepalm: An alternative is to fetch a specific git commit. They describe how to do it in the [spack tutorial](https://spack.readthedocs.io/en/latest/packaging_guide.html#git)

benchmarks/spack/repo/packages/arepo/package.py
```diff
@@ -0,0 +1,59 @@
...
+    homepage = "https://gitlab.mpcdf.mpg.de/vrs/arepo.git"
+    git = "https://gitlab.mpcdf.mpg.de/vrs/arepo.git"
+    
+    
+    version("master", branch="master")
+   
+
+    depends_on('mpi')
+    depends_on('gsl')
+    depends_on('hdf5')
```
tkoskela, line 16: I had to add this dependency to build on my local machine
```suggestion
    depends_on('mpi')
    depends_on('gsl')
    depends_on('hdf5')
    depends_on('gmp')
```

tkoskela (changes requested):
Hi Gokmen. Sorry I did not notice this PR was open. I need to adjust my notification settings.

A couple of quick comments.

benchmarks/spack/repo/packages/arepo/package.py
```diff
@@ -0,0 +1,59 @@
...
+
+    variant('fftw', default=False, description='FFTW support')
+    variant('hdf5', default=True, description='HDF5 support')
+    variant('hwloc', default=False, description='HWLOC support')
+        
+    depends_on('fftw',  when='+fftw')
+    depends_on('hwloc', when='+hwloc')
+    depends_on("mpich", when='^mpich')
+    depends_on("openmpi", when='^openmpi')
```
tkoskela, line 25: Bit of a silly question, but isn't spack smart enough to figure this out on its own?

benchmarks/spack/repo/packages/arepo/package.py
```diff
@@ -0,0 +1,59 @@
...
+
+        mkdirp(prefix.bin)
+        install('./Arepo', prefix.bin)
+        
+
+    
+        
+
+
+
```
tkoskela, line 58:
```suggestion
```

benchmarks/spack/repo/packages/arepo/package.py
giordano, line 16: Should hdf5 be an optional dep added only when hdf5 support is requested?

benchmarks/spack/repo/packages/arepo/package.py
gokmenkilic: Thanks for your suggestion! I have updated the Spack recipe to simplify the MPI dependency handling

benchmarks/spack/repo/packages/arepo/package.py
gokmenkilic: Exactly, I have already added it as a variant.

benchmarks/apps/arepo/arepo.py
gokmenkilic: I was confused at the beginning not to see any tags and version for Arepo and now it makes sense! Thank you.

benchmarks/spack/repo/packages/arepo/package.py
gokmenkilic: I have added the `depends_on('gmp')`

gokmenkilic:
> Hi Gokmen. Sorry I did not notice this PR was open. I need to adjust my notification settings.
> 
> A couple of quick comments.

Hi Thomas,

It is fine. I made the changes

benchmarks/spack/repo/packages/arepo/package.py
tkoskela: I think you also need to add something like
```
depends_on('hdf5', when='+hdf5')
```

benchmarks/spack/repo/packages/arepo/package.py
```diff
@@ -0,0 +1,47 @@
...
+        copy('Template-Makefile.systype', 'Makefile.systype')
+        copy('Template-Config.sh', 'Config.sh')
+        
+        specific_config_path = 'examples/yee_2d/Config.sh'
+            
+        self.config_file = specific_config_path  
+
+        with open('Makefile.systype', 'w') as f:
+            f.write('SYSTYPE="Darwin"\n')
+
```
giordano, line 34: Is this supposed to be always set to Darwin, or only on macOS?

gokmenkilic: I am not quite sure but here is two way to do that:

The first thing after obtaining a local copy of the code is to set a SYSTYPE variable, either by typing:

`export SYSYTPE="MySystype"`
(assuming you run the bash-shell) or by copying the Template-Makefile.systype file:

`cp Template-Makefile.systype Makefile.systype`
and uncommenting the intended SYSTYPE variable in Makefile.systype.

benchmarks/apps/arepo/arepo.py
ilectra, line 15:
```suggestion
    spack_spec = 'arepo@2024-04-01'
```

benchmarks/spack/repo/packages/arepo/package.py
tkoskela: I think the question here was whether always seting this to `Darwin` is correct. On my Ubuntu laptop I need to manually change it to `Ubuntu` to get the include paths right, which is not ideal. 

I would have assumed that you could just leave `SYSTYPE` undefined and spack would provide the right paths to the dependencies, but this does not seem to be the case. I didn't really dig deeper into the build system to find out what it's doing.

benchmarks/spack/repo/packages/arepo/package.py
```diff
@@ -0,0 +1,49 @@
...
+   
+    depends_on('mpi')
+    depends_on('gsl')
+    depends_on('gmp')
+    depends_on('hdf5', when='+hdf5')    
+
+
+    variant('fftw', default=False, description='FFTW support')
+    variant('hdf5', default=True, description='HDF5 support')
+    variant('hwloc', default=False, description='HWLOC support')
```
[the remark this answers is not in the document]
ilectra: See just below: 
```
depends_on('fftw',  when='+fftw')
```
It worked for me, default and variants.

benchmarks/spack/repo/packages/arepo/package.py
ilectra: This is definitely strange: I'd expect it to fail on a non-mac, which is what Tuomas is seeing, but it didn't fail for me on cosma. :thinking:

tkoskela: Perhaps on cosma when you have the right modules loaded you don't need to set include paths in the makefile and the `SYSTYPE` doesn't matter. I would like the build system to work without hard-coded paths, that would make life a lot easier.

gokmenkilic: I agree with @tkoskela, it worth to escape hard-coded paths. They could cause mess for general perspective.

benchmarks/spack/repo/packages/arepo/package.py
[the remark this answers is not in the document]
ilectra: Oh... Do we even want to build without fftw? The (very divergent) branch I'm looking at needs it by default. (I'll add a variant for that in another PR, once we merge this)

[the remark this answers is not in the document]
tkoskela: Unless I'm misunderstanding what `'FFTW Support'` means here?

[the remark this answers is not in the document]
tkoskela: Any thoughts on this?

connoraird:
I am now hitting issue with running benchmarks on cosma8. I think the fixes in this PR could help. Can we split this out into it's own PR and merge?

giordano:
> Can we split this out into it's own PR and merge?

That sounds sensible.

PR #333, opened 2024-06-21
Add ReFrame config for CSD3 Ponte Vecchio (Dawn) partition

tkoskela:
I've ran several benchmarks (Stream, Sombrero, Babelstream) successfully with this. Does it look ok?

I haven't included a spack env because Kacper recommended chaining to the Dawn spack instance. There is a script `/usr/local/dawn/software/spack/spack-tools/bin/spack-dawn-upstream.sh` which does it. I think this is in general a better idea than maintaining our own spack envs. Ideas welcome how to do it in a sustainable way.

### DIFF
--- a/benchmarks/reframe_config.py
+++ b/benchmarks/reframe_config.py
@@ -125,7 +125,7 @@ site_configuration = {
                     'name': 'pvc',
                     'descr': 'Ponte Vecchio (Dawn) compute nodes',
                     'scheduler': 'slurm',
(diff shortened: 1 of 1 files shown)


giordano (commented):
I don't have access, so I can't test it, but looks overall good to me.  Anything to add to the docs?  Since this is just a new partition of an existing system maybe not?

giordano (approved):
I fatfingered the button, I wanted to approve rather than simply commenting!

PR #335, opened 2024-07-18
Require numpy 1 due to incompatibility with Streamlit

### LINKED ISSUE #334: Post-processing breaks with numpy/2.0.0
```
Traceback (most recent call last):
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 589, in _run_script
    exec(code, module.__dict__)
  File "/home/tkoskela/git_repos/excalibur-tests/post-processing/streamlit_post_processing.py", line 7, in <module>
    from post_processing import PostProcessing
  File "/home/tkoskela/git_repos/excalibur-tests/post-processing/post_processing.py", line 11, in <module>
    from plot_handler import plot_generic
  File "/home/tkoskela/git_repos/excalibur-tests/post-processing/plot_handler.py", line 7, in <module>
    from bokeh.models import HoverTool, Legend
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/bokeh/models/__init__.py", line 33, in <module>
    from ..core.property.dataspec import expr, field, value  # Legacy API
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/bokeh/core/property/dataspec.py", line 35, in <module>
    from .color import Color
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/bokeh/core/property/color.py", line 30, in <module>
    from .container import Tuple
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/bokeh/core/property/container.py", line 38, in <module>
    from .enum import Enum
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/bokeh/core/property/enum.py", line 27, in <module>
    from .primitive import String
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/bokeh/core/property/primitive.py", line 37, in <module>
    bokeh_bool_types += (np.bool8,)
  File "/home/tkoskela/python_envs/lexci/lib/python3.10/site-packages/numpy/__init__.py", line 410, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'bool8'
```

tkoskela:
Update bokeh version to work with latest numpy

Closes #334

### DIFF
--- a/post-processing/test_post_processing.py
+++ b/post-processing/test_post_processing.py
@@ -172,7 +172,7 @@ def test_read_perflog(run_sombrero):
     # get dataframe from complete perflog
     df = log_hand.read_perflog(sombrero_log_path)
 
-    EXPECTED_FIELDS = ["job_completion_time", "version", "info", "jobid", "num_tasks",
+    EXPECTED_FIELDS = ["job_completion_time", "reframe version", "info", "jobid", "num_tasks",
                        "num_cpus_per_task", "num_tasks_per_node", "num_gpus_per_node",
                        "flops_value", "flops_unit", "flops_ref", "flops_lower_thres",
                        "flops_upper_thres", "spack_spec", "test_name", "tasks", "cpus_per_task",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 dependencies = [
-    "reframe-hpc >= 4.3.0, < 5.0.0",
+    "reframe-hpc >= 4.6.1, < 5.0.0",
     "matplotlib >= 3.0.0",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ test = [
 ]
 post-processing = [
     "pandas >= 2.0.1",
-    "bokeh == 2.4.3",
+    "bokeh == 3.1.0",
     "titlecase >= 2.4.1",
     "streamlit >= 1.30.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,9 +34,10 @@ test = [
 ]
 post-processing = [
     "pandas >= 2.0.1",
-    "bokeh >= 3.1.0",
+    "bokeh == 2.4.3",
     "titlecase >= 2.4.1",
     "streamlit >= 1.30.0",
+    "numpy < 2.0.0",
 ]
 
 [tool.setuptools_scm]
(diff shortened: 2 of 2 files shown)


giordano (commented):
Looks good to me, but I don't know why we were pinning an exact (and quite old) version.  Latest version at the moment is 3.5.0.

tkoskela:
I actually meant to use `>=` not `==`, let me fix

tkoskela:
Ah, there was a reason for that specific version. This could be a problem
```
StreamlitAPIException: Streamlit only supports Bokeh version 2.4.3, but you have version 3.5.0 installed. Please run pip install --force-reinstall --no-deps bokeh==2.4.3 to install the correct version.
```

pineapple-cat (commented):
Yes, we need to use Bokeh v2.4.3 so it doesn't break Streamlit. Can we just cap the numpy version below 2.0.0 for now?

pineapple-cat (commented):
I think the test failure that's happening now is the same one as in the line plot PR, but I can't quite tell what's causing it at a glance. Maybe there's another library we need to version cap – pandas perhaps?

tkoskela:
It's failing on
```
    assert all(c in EXPECTED_FIELDS for c in df.columns.tolist())
```
On a quick look through the perflog the fields look ok. Maybe something has changed in the spac spec dict?

pineapple-cat:
If it helps, I think the change causing this may have taken place 28 - 29 May. Using the line plot PR as reference again, all tests were passing on the 28th, but when I pushed some inconsequential changes on the 29th this specific test started failing.

tkoskela:
The field `version` has changed to `reframe version` (I assume in a new version of ReFrame). I'm running `4.6.2`. I updated it in the test and now it passes. I noticed we don't have whitespace in other field names. Is that going to cause any issues @pineapple-cat?

pineapple-cat:
Of course, I forgot about ReFrame field name updates; thank you for finding the issue.

pineapple-cat (approved):
The presence of whitespace in "reframe version" bothers me on a psychological level, but I don't think there's anything in our config handler that explicitly can't handle it. We may want to replace all whitespace with an underscore to ensure consistency in label handling for later stages, but it's not strictly necessary. I just ran some tests with whitespace in field names and the resulting graphs look fine.

tkoskela:
This change seems to have happened in ReFrame `4.6.1`. Are we happy to update the requirement to that?

giordano:
I think so.  Better to be explicit about the actually supported versions.

PR #342, opened 2024-09-09
Some changes to `spack_root_to_path`

giordano:
* Add comments to improve clarity of all the conditions considered
* There were unnecessary `else` branches
* Cater for the dramatic case where both `SPACK_ROOT` and `PATH` are not set
* Fix the case where `PATH` isn't set but `SPACK_ROOT` is (we were returning the function `dir` before, fun)
* Fix the case where `spack_bindir` isn't in `PATH` (we were using `*` for non-commutative concatenation of strings, it turns out in Python one should use `+` instead for this operation)

### DIFF
--- a/benchmarks/reframe_config.py
+++ b/benchmarks/reframe_config.py
@@ -18,16 +18,23 @@ def spack_root_to_path():
     spack_root = os.getenv('SPACK_ROOT')
     path = os.getenv('PATH')
     if spack_root is None:
+        # Somehow we don't know what's the Spack root, then return PATH as is,
+        # but if also PATH is not set (what a dramatic case) then return an
+        # empty string.
+        return "" if path is None else path
+
+    spack_bindir = os.path.join(spack_root, 'bin')
+    if path is None:
+        # Somehow PATH isn't set, only return `spack_bindir`
+        return spack_bindir
+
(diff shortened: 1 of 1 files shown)


benchmarks/reframe_config.py
```diff
@@ -18,16 +18,23 @@ def spack_root_to_path():
...
+        return "" if path is None else path
+
+    spack_bindir = os.path.join(spack_root, 'bin')
+    if path is None:
+        # Somehow PATH isn't set, only return `spack_bindir`
+        return spack_bindir
+
+    if spack_bindir in path.split(os.path.pathsep):
+        # `spack_bindir` is already in PATH, return the environment
+        # variable as is.
```
tkoskela, line 33: To make this more readable, could you start with the most likely cases and leave the corner cases to the end?

giordano: I first need to check `spack_bindir` and `path` aren't `None` though, which is what the two `if`s above are doing.

tkoskela: Right, I didn't realise all the if branches `return` and thought they'd all get checked regardless. It's fine then

tkoskela (approved):

Issue #347, opened 2024-10-23
Support running tracing profiler with the application


giordano:
We're planning to support running the applications with tracing profilers, which are generally slightly easier to deal with than those using performance counters (like [linux perf](https://perfwiki.github.io/main/), or [likwid](https://github.com/RRZE-HPC/likwid)).

Some of the profilers we're considering supporting:
* [ ] [Intel Vtune](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html): proprietary, but can be used free of charge, only supports x86 architectures
* [ ] [Nvidia Nsight Compute](https://developer.nvidia.com/nsight-compute): proprietary, but can be used free of charge, supports x86 and aarch64 architectures, currently not available in Spack, supports also MPI and GPU profiling
* [ ] [Extrae](https://tools.bsc.es/extrae): free and open source, available on a wide range of platforms, less popular than above options, supports also MPI and GPU profiling, but not extremely simple to use (the visualisation counterpart, [Paraver](https://tools.bsc.es/paraver), is also powerful but not intuitive to use)
* [ ] [Score-P](https://www.vi-hps.org/projects/score-p/): free and open-source, but I'm not familiar with it myself, can't comment more about it

ilectra:
How about Linaro Forge?  (used to be ArmForge)

giordano:
Unlike all profilers mentioned above, linaro requires a license though, which makes it much harder to test.

ilectra:
Sure, but we do have a licence on (at least!) DiRAC systems.

tkoskela:
The license is paid by ExCALIBUR so it might be gone after March :imp: I can ask about this in the next H&ES meeting.

giordano:
Note to self: in ReFrame for profilers which need to wrap the `mpirun` call we can use [`LauncherWrapper`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.launchers.LauncherWrapper)

PR #348, opened 2024-10-28
Automatically run a profiler with the executable

giordano:
Proof-of-concept for #347, at the moment only supporting vtune.  Everything is up-to-discussion, this is just to show how this could work. 

You can run this for the stream example with
```
reframe -c benchmarks/examples/stream -r --performance-report -S profiler=vtune
```
(the main point is the option `-S profiler=vtune`, the rest are whatever other options you want to pass to `reframe`).


tkoskela (approved):
LGTM!

PR #356, opened 2025-01-17
Updates to Isambard-3 spack env and purify benchmark

tkoskela:
Updates purify benchmark to use the latest 5.0 version, adds new benchmarks for the ForwardBackward algorithm and add some sanity checks to parameters.

Adds python packages to isambard-3 spack env

I've commented out external cray-fftw because of the fftw bug that breaks cmake builds. The bug is patched in the spack package so building your own fftw works just fine.


benchmarks/spack/isambard3/grace/spack.yaml
```diff
@@ -25,6 +23,18 @@ spack:
...
       extra_rpaths: []
   packages:
+    cmake:
+      externals:
+      - spec: cmake@3.20.4
+        prefix: /usr
+    python:
+      externals:
+      - spec: python@3.11.7+bz2+crypt+ctypes+dbm+lzma+pyexpat+pythoncmd+readline+sqlite3+ssl+tix+tkinter+uuid+zlib
+        prefix: /lfs1i3/home/lexci-i3/tkoskela.lexci-i3/python_envs/test
```
giordano, line 33: This is in your home
```suggestion
```

tkoskela: Oops. I see, it had picked up the python from my virtual environment. Well spotted!

benchmarks/spack/isambard3/grace/spack.yaml
```diff
@@ -146,9 +152,25 @@ spack:
...
+        - cray-python
+        prefix: /opt/cray/pe/python/3.11.7
+      buildable: false
+    py-mpi4py:
+      externals:
+      - spec: py-mpi4py@3.1.4
+        modules:
+        - cray-python
+        prefix: /opt/cray/pe/python/3.11.7
+      buildable: false
```
giordano, line 173: Any reason for using `buildable: false` for this packages?  Also, I believe you want also to append `^ python@3.11.7` to all these specs, to avoid making it match with other versions of python.

tkoskela: I thought `buildable: false` guarantees that the external will always get used. Is it too strong requirement? Happy to remove it if you think that's better. Adding the python version to the spec

giordano: `buildable: false` complicates environment resolution quite a bit whenever a package appears in the environment.  I'd expect troubles for `py-numpy` and `py-scipy` because they're somewhat common, `py-mpi4py` less so and I'd hope spack links it to the right MPI library anyway.

giordano (approved):

PR #357, opened 2025-01-28
Add ziatest benchmark

giordano:
Still to do:

* [ ] support [microsecond printing](https://gitlab.com/NERSC/N10-benchmarks/ziatest/-/blob/523122f4907f32f735d8aff09c4b801da12066b5/ziaprobe.c#L187-L189)
* [ ] let users set number of tasks, nodes, etc from the command line


benchmarks/spack/repo/packages/ziatest/package.py
giordano, line 1: I plan to upstream this in Spack, but I'm keeping this here for the time being while I iterate to make sure I don't need more patches (I have at least 2 more to improve the code, but aren't strictly necessary here)

giordano:
This is ready to go from my point of view.

benchmarks/apps/ziatest/ziatest.py
```diff
@@ -0,0 +1,50 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+from benchmarks.modules.utils import SpackTest
+
+
+@rfm.simple_test
+class ZiatestBenchmark(SpackTest):
+    valid_systems = ['-gpu']
```
giordano, line 8: I'm not sure about this one, I guess people may want to run the benchmark on any type of system.

PR #360, opened 2025-02-10
[spack] Update environment for GitHub Actions

### LINKED ISSUE #359: Update spack env for GitHub actions
GHA has updated `ubuntu-latest` to `24.04`. We need to update the compilers for spack to be compatible.

giordano:
I'm a bit torn about how to best fix #359.  I tried [automatically populating the Spack environment](https://github.com/ukri-excalibur/excalibur-tests/commit/4631503010a1389d8b21400335c4b2e9fa468b7c) in every job, which should hopefully always ensure consistency, but [that step can take over a minute](https://github.com/ukri-excalibur/excalibur-tests/actions/runs/13244795136/job/36968533449#step:9:1), depending on the mood of the machine. The solution proposed here (just manually update the environment) is less future-proof, but would avoid wasting time in every job.  I'm happy to change to the other solution if considered preferable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-latest
+          - ubuntu-24.04
         spack_version:
           # In README.md we claim compatibility with v0.20.0
           - v0.20.0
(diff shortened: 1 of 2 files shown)


tkoskela (approved):
I am happy with manually updating the versions for now. At least you've also fixed the Ubuntu version so this will not sneak up on us again in two years :sweat_smile:

PR #363, opened 2025-02-21
[CSD3] Update config and Spack environments following OS upgrade

### LINKED ISSUE #355: Update CSD3 Cascade Lake Spack environment
In July 2024, we retired CentOS 7 (as it went end-of-life) and upgraded our Cascade Lake nodes to Rocky Linux 8. A new software stack has been built for this partition and OS. The Spack environment therefore needs to be updated accordingly.

I don't have the time the make these changes at the moment but I thought I would at least open this issue.

mirenradia:
In July 2024, the nodes in the Cascade Lake partition were updated from CentOS 7 to Rocky Linux 8. They now share the same OS as other CPU partitions on CSD3 so this combines all the partitions under one system again rather than having separate ones for different OSes.

We switch to using the latest available software stack which is that built for the Cascade Lakes. This works on all CPU partitions. Instead of adding system provided packages as externals, we instead chain our Spack environment to that of the system Spack instance.

This resolves #355.


benchmarks/spack/csd3/upstreams.yaml
giordano, line 1: Ah, we never used that so far, but it's a good idea if it works!

giordano (approved):
Not tested myself, but looks good overall!  And curious to see how the `upstream` works out!

Note that there is a merge conflict now.

mirenradia:
I've rebased onto `main` so there shouldn't be any conflicts now.

PR #365, opened 2025-03-06
[kathleen] Add `+generic-names` to `intel-oneapi-mpi`

giordano:
Compiler wrappers use generic names, let Spack know it so that it can find them correctly.


connoraird (approved):
This change was required for me so looks good

Issue #336, opened 2024-07-18
Add the option for logarithmic scaling to axes in post-processing graphs


pineapple-cat:
Should be relatively simple to add this as a config setting that can be switched on and off for each numeric axis.

PR #369, opened 2025-03-19
Add numeric axis log scaling options for post-processing

pineapple-cat:
Addresses #336.

### DIFF
--- a/post-processing/post_processing.py
+++ b/post-processing/post_processing.py
@@ -66,10 +66,10 @@ class PostProcessing:
             config.x_axis["value"], config.y_axis["value"], *config.get_y_scaling(), config.series_filters)
         # log axes
         if (config.x_axis.get("logarithmic") and
-            pd.api.types.is_numeric_dtype(self.df[config.x_axis["value"]].dtype)):
+            pd.api.types.is_float_dtype(self.df[config.x_axis["value"]].dtype)):
             self.df[config.x_axis["value"]] = np.log10(self.df[config.x_axis["value"]])
         if (config.y_axis.get("logarithmic") and
-            pd.api.types.is_numeric_dtype(self.df[config.y_axis["value"]].dtype)):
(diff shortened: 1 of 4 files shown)


post-processing/post_processing.py
```diff
@@ -63,6 +64,13 @@ def run_post_processing(self, config: ConfigHandler):
         # scale y-axis
         self.transform_df_data(
             config.x_axis["value"], config.y_axis["value"], *config.get_y_scaling(), config.series_filters)
+        # log axes
+        if (config.x_axis.get("logarithmic") and
+            pd.api.types.is_numeric_dtype(self.df[config.x_axis["value"]].dtype)):
+            self.df[config.x_axis["value"]] = np.log(self.df[config.x_axis["value"]])
+        if (config.y_axis.get("logarithmic") and
+            pd.api.types.is_numeric_dtype(self.df[config.y_axis["value"]].dtype)):
+            self.df[config.y_axis["value"]] = np.log(self.df[config.y_axis["value"]])
```
pineapple-cat, line 73: Should change the log function used to `np.log10`.

pineapple-cat:
I can't find a way to use `y_axis_type="log"` without breaking the plot in some way, regardless of whether Bokeh 2.x or 3.x is used, and trying to log int columns comes with its own set of headaches. If we want to be able to transform (log or scale) int columns, this should be made into a separate issue.

Also, now that Streamlit has Bokeh 3.x support I'll make a separate upgrade PR.

pineapple-cat:
Unless we want to add additional information about log axes to the README, the example config change was so small I just ended up including it with the docs update in #374.

ilectra:
> Unless we want to add additional information about log axes to the README, the example config change was so small I just ended up including it with the docs update in #374.

Yes, I saw that :) . Add it to the README as well, it's nice to advertise new functionality to the users.

PR #370, opened 2025-03-21
Gromacs benchmark

connoraird:
- Fixing some Spack machine config for cosma8, kathleen and myriad.
- Adding benchmarks for gromacs with CPU and GPU


benchmarks/apps/gromacs/README.md
giordano, line 1: Can you please add a little bit more information about the benchmark and how to run it?  You can for example take [CP2K](https://github.com/ukri-excalibur/excalibur-tests/blob/f879d83145fbac6e0e156a3193c10cdb8a1e9bdc/benchmarks/apps/cp2k/README.md) for inspiration.  Also add the application entry under https://github.com/ukri-excalibur/excalibur-tests/blob/f879d83145fbac6e0e156a3193c10cdb8a1e9bdc/mkdocs.yml#L23

benchmarks/spack/cosma8/compute-node/spack.yaml
```diff
@@ -321,16 +321,8 @@ spack:
...
-      - spec: cmake@2.8.12.2
-        prefix: /usr
-      - spec: cmake@3.11.4
-        prefix: /cosma/local/cmake/3.11.4/
-      - spec: cmake@3.18.1
-        prefix: /cosma/local/cmake/3.18.1/
-      - spec: cmake@3.25.1
-        prefix: /cosma/local/cmake/3.25.1/
-      - spec: cmake@3.28.3
-        prefix: /cosma/local/cmake/3.28.3/
```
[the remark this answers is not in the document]
connoraird: These removed versions fail as they do not have libssl linked. For example
```
$ ldd /cosma/local/cmake/3.28.3/bin/cmake
	linux-vdso.so.1 (0x00007ffcfdbbb000)
	libssl.so.10 => not found
	libcrypto.so.10 => not found
	libdl.so.2 => /lib64/libdl.so.2 (0x000014a6e1283000)
	librt.so.1 => /lib64/librt.so.1 (0x000014a6e127e000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x000014a6e1279000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x000014a6e1000000)
	libm.so.6 => /lib64/libm.so.6 (0x000014a6e0f25000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x000014a6e125c000)
	libc.so.6 => /lib64/libc.so.6 (0x000014a6e0c00000)
	/lib64/ld-linux-x86-64.so.2 (0x000014a6e129d000)
```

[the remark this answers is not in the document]
giordano: Ah, I guess they may have updated openssl without rebuilding those packages.

benchmarks/apps/gromacs/gromacs.py
```diff
@@ -0,0 +1,98 @@
...
+    @run_before('performance')
+    def set_test_perf_patterns(self):
+        """Set the regex performance pattern to locate"""
+        self.perf_patterns = {
+            'Rate': sn.extractsingle(r'Performance:\s+(?P<rate>\S+)(\s+\S+){1}',
+                                     self.expected_output_file, 'rate', float),
+            'Energy': sn.extractsingle(r'\s+Potential\s+Kinetic En\.\s+Total Energy\s+Conserved En\.\s+Temperature\n'
+                                       r'(\s+\S+){2}\s+(?P<energy>\S+)(\s+\S+){2}\n'
+                                       r'\s+Pressure \(bar\)\s+Constr\. rmsd',
+                                       self.expected_output_file, 'energy', float),
```
tkoskela, line 72: Is this a performance metric?

benchmarks/apps/gromacs/gromacs.py
```diff
@@ -0,0 +1,98 @@
...
+    tags = {"cpu"}
+
+    executable_opts = ['mdrun', '-noconfout', '-dlb', 'yes', '-s', input_data_file]
+    num_nodes_param = parameter([1, 2, 3, 4])
+
+
+@rfm.simple_test
+class StrongScalingSpackGPU(GROMACSBenchmark):
+    spack_spec = 'gromacs@2024 +mpi+cuda'
+    tags = {"gpu"}
```
tkoskela, line 87: Tags are good to have, but shouldn't you use the `valid_systems` variable so that the GPU benchmark won't run on a system without a GPU and vice versa? For example you can have a look at babelstream https://github.com/ukri-excalibur/excalibur-tests/blob/f095234cd131161a5324ed374fcc0ad785807dbb/benchmarks/apps/babelstream/babelstream.py#L60-L74

benchmarks/apps/gromacs/gromacs.py
tkoskela: If you're using it for sanity checking, then you should extract it as a sanity pattern

benchmarks/spack/cosma8/compute-node/spack.yaml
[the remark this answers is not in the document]
gokmenkilic: I had the same issue with Swift Benchmark. I think in Rocky system you do not need the specify explicitly cmake. With removing all cmake versions works for me.

tkoskela (approved):
I haven't tested running this, but the changes you made look good to me

PR #371, opened 2025-03-21
Hemepure benchmark

connoraird:
- This relies on some changes from #370 in order to work on cosma8, kathleen or myriad
- Adding benchmarks for hememure on CPU and GPU


benchmarks/apps/hemepure/hemepure.py
```diff
@@ -0,0 +1,83 @@
...
+            'SIMULATION FINISHED', self.stdout
+        )
+
+    @run_before('performance')
+    def set_test_perf_patterns(self):
+        """Set the regex performance pattern to locate"""
+        self.perf_patterns = {
+            'Runtime': sn.extractsingle(r'\[Rank \d+, (?P<runtime>\S+) s, \d+ kB] :: SIMULATION FINISHED',
+                                     self.stdout, 'runtime', float, item=-1)
+        }
```
tkoskela, line 60: When I asked Jon McCulloug about performance metrics, his response was

> Regarding performance metrics, time steps / wall clock does make sense, though for LBM codes this is usually expressed in terms of millions of lattice updates per second (MLUPS)
Calculated as (number sites * number steps)/(1000000 * wall clock time for simulation). We typically separated initialisation time and simulation time for this. MLUPS can also be expressed cumulatively or on a per core or per node basis depending on the information you want to convey.

I haven't looked at the output so I don't know what's there. If we can extract either time steps/unit of wall time or MLUPS, those would be a better performance metric

benchmarks/apps/hemepure/README.md
```diff
@@ -0,0 +1,49 @@
+# Hemepure benchmark
+
+HemeLB is a high performance lattice-Boltzmann solver optimized for simulating blood flow through sparse geometries, such as those found in the human vasculature. It is routinely deployed on powerful supercomputers, scaling to hundreds of thousands of cores even for complex geometries. HemeLB has traditionally been used to model cerebral blood flow and vascular remodelling in retinas, but is now being applied to simulating the fully coupled human arterial and venous trees.
+
+HemePure is a optimized verion of HemeLB with improved memory, compilation and scaling
+
+This directory includes a data directory, `input_data/pipe`, which can be used to benchmark both GPU and CPU machines. This data file was taken from https://github.com/UCL-CCS/HemePure/tree/master/cases/pipe.
```
tkoskela, line 7: Is this the same as the TestPipe data in https://zenodo.org/records/14859634

connoraird: It does not appear to be the same

benchmarks/apps/hemepure/hemepure.py
connoraird: Okay so simulated time steps divided by reported the wall time?

PR #374, opened 2025-04-01
Update post-processing to use streamlit-bokeh

### LINKED ISSUE #330: Fix scaling in Streamlit post-processing
ln 323 in streamlit_post_processing.py was recently changed `"x_value": y_scaling_x` → `"x_value": str(y_scaling_x)` to fix a config export issue, however this broke scaling functionality in the web app.

We could try to fix this by typing the scaling x-value before performing comparisons with it, such as:
- changing ln 264 in post_processing.py `scaling_x_value_mask = (self.df[x_column] == scaling_column.get("x_value"))` → `scaling_x_value_mask = (self.df[x_column] == self.val_as_col_dtype(scaling_column["x_value"], x_column).iloc[0])`
- under ln 251 in streamlit_post_processing.py adding and using `typed_x_val = state.post.val_as_col_dtype(axis["scaling"]["column"]["x_value"], state.x_axis_column).iloc[0]`

However, these fixes alone are not enough because we must find a way to get around the error thrown if the app tries to type a "None" string as numeric (specifically, I found this was an issue when typing to Int64).

### LINKED ISSUE #345: Empty line in dataframe
My LEXCI benchmarks produce an odd empty line in my data frame

image

The [raw perflog](https://github.com/user-attachments/files/17095609/PurifyMOBenchmark_PratleyEtAl.log) only has 8 lines with `%tasks=1` so I'm not sure where this is coming from. I also get it in the [csv export](https://github.com/user-attachments/files/17095604/2024-09-23T09-53_export.csv). Not a blocker since I filter it away anyways, but maybe worth having a look at.

pineapple-cat:
Upgrades Bokeh and Streamlit versions + fixes some miscellaneous bugs.

- [x] Update post-processing to plot with `streamlit-bokeh` so we can use newer versions of Bokeh and Streamlit.
- [x] Fix plotting issues when y-axis is typed as int.
- [x] Fix Streamlit Session State API warning upon config upload.
- [x] Fix #330.
- [x] Fix #345.

### DIFF
--- a/post-processing/streamlit_post_processing.py
+++ b/post-processing/streamlit_post_processing.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import traceback
 from pathlib import Path
 
@@ -355,7 +356,8 @@ def scaling_select(axis: dict):
                      key="y_axis_scaling_x_value")
 
     if "y_axis_custom_scaling_val" not in state:
-        state["y_axis_custom_scaling_val"] = axis["scaling"].get("custom") if axis.get("scaling") else None
+        state["y_axis_custom_scaling_val"] = (str(axis["scaling"].get("custom")) if axis["scaling"].get("custom")
+                                              else None if axis.get("scaling") else None)
     st.text_input("custom scaling value", placeholder="None", key="y_axis_custom_scaling_val",
                   help="Assign a scaling value that isn't in the data. Will clear all other scaling selections.")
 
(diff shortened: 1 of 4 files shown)


post-processing/streamlit_post_processing.py
```diff
@@ -102,12 +111,77 @@ def update_config():
 
     state = st.session_state
     uploaded_config = state.uploaded_config
+    df = state.post.df
+
     if uploaded_config:
         try:
             config_dict = load_config(uploaded_config)
             state.config = ConfigHandler(config_dict)
+            config = state.config
```
ilectra, line 120: Is all this needed to solve https://github.com/ukri-excalibur/excalibur-tests/issues/345#issuecomment-2517661658 ? How come with streamlit version 1.30.0 it works?

ilectra (commented):
There's a lot of changes here I cannot claim I understand. But I'll trust you that they work and they improve the code :) 
Just one question - see comment.

post-processing/streamlit_post_processing.py
pineapple-cat: Yes, these `update_config` changes solve the exact problem you're describing in that comment. My guess is that in the newer Streamlit versions they must have either changed something about how session state updates or how text input / selectboxes update their values.

It used to be the case that changing the config in the session state was enough to update everything that pulled from it, but now we get a completely different behaviour where the text input and selectbox fields don't update according to the new config and instead change some of the config values back to whatever was originally in the UI fields. I suspect it has something to do with allowing empty / null / None in those fields, as type selectboxes don't seem to have this issue. Either way, it should all be sorted now.

ilectra: Ok, I see. Is it worth it to open an issue in the upstream streamlit repo? I mean, they degraded in functionality...

pineapple-cat: I agree, it was quite tedious to have to write in these additions to avoid breaking the UI when it just worked before, but I don't know why they made the changes they did, or even necessarily what the changes were, so not sure I'd like to open an issue without at least investigating the change logs.

ilectra: Sure, check the logs first. But opening an issue doesn't mean you have to solve it as well, you're doing them a favour by spotting something they didn't.

PR #366, opened 2025-03-13
Fix arepo to run on cosma.

tkoskela:
Cherry-picked from #328 to close #339 
- Fix OS name in spack environment for cosma for new OS after upgrade
- Add some new packages and compilers to spack environment for cosma post-upgrade (they may or may not be needed for AREPO...)

PR #378, opened 2025-04-10
Updated spack.yaml for Rocky System

gokmenkilic:
Address #366 

Removed old compilers which break SwiftBenchmark. This version works fine on cosma8.

### DIFF
--- a/benchmarks/spack/cosma8/compute-node/spack.yaml
+++ b/benchmarks/spack/cosma8/compute-node/spack.yaml
@@ -1,143 +1,9 @@
-# This is a Spack Environment file.
-#
-# It describes a set of packages to be installed, along with
-# configuration settings.
 spack:
-  # add package specs to the `specs` list
   specs: []
   view: false
   include:
   - ../../common.yaml
   compilers:
-  - compiler:
-      spec: gcc@4.8.2
-      paths:
-        cc: /cosma/local/aocc/AOCC-1.3.0-Fortran-Prerequisites/gcc_4.8.2_install/bin/gcc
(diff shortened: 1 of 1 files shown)


benchmarks/spack/cosma8/compute-node/spack.yaml
```diff
@@ -281,19 +150,6 @@ spack:
...
-        cc: /usr/bin/gcc
-        cxx: /usr/bin/g++
-        f77: /usr/bin/gfortran
-        fc: /usr/bin/gfortran
-      flags: {}
-      operating_system: rocky9
-      target: x86_64
-      modules: []
-      environment: {}
-      extra_rpaths: []
```
giordano, line 296: This compiler doesn't exist anymore?

benchmarks/spack/cosma8/compute-node/spack.yaml
```diff
@@ -593,4 +372,4 @@ spack:
       - spec: xz@5.2.5
         prefix: /cosma/local/intel/oneAPI_2021.1.8/intelpython/latest
       - spec: xz@5.2.2
-        prefix: /usr
+        prefix: /usr
```
giordano, line 375: Missing newline (which was there before this change):
```suggestion
        prefix: /usr

```

gokmenkilic (commented):
Removed old compilers and added new gcc compilers. Having cmake compilers into the Cosma8 spack file giving issues. In Rocky system I believe it will pick automatically.

benchmarks/spack/cosma8/compute-node/spack.yaml
gokmenkilic: Now,it is okay I guess.

giordano (approved):

Issue #381, opened 2025-05-14
Investigate series requirements for post-processing


pineapple-cat:
Currently, if series are present in a config, there must be at least 2 conditions in the list. This requirement should be revisited, ideally with some updated tests, to check whether it is necessary.

Issue #380, opened 2025-05-14
Disable generic x-axis log scaling


pineapple-cat:
Applying the log scale to a generic x-axis only serves to make the labels less readable, so there's no particular reason this should be allowed.

PR #382, opened 2025-05-27
Miscellaneous post-processing UI fixes

pineapple-cat:
- [x] #381 
- [x] #380
- [x] Fix default minimum range to fit all data.
- [x] Prevent session state warning when updating ranges.
- [x] Add vertical generic x-axis labels.


post-processing/config_handler.py
```diff
@@ -297,8 +297,6 @@ def read_config(config: dict):
 
     # check optional series information
     if config.get("series"):
-        if len(config.get("series")) == 1:
-            raise RuntimeError("Number of series must be >= 2.")
```
pineapple-cat, line 301: I'm under the impression that this was a cosmetic decision when first added.

PR #392, opened 2026-05-19
Fix python deps

owainkenwayucl:
Installing the package misses a bunch of dependencies, including pandas (missing even in the base version of the ExCALIBUR benchmarks), but also the atlassian and openstack libraries.

It's possible there are others?


owainkenwayucl:
Sorry - pull request against wrong branch

